PR #18, opened 2016-04-04
t/17: Implement the themes architecture

### LINKED ISSUE #17: Implement the themes architecture
An UI framework part of https://github.com/ckeditor/ckeditor5/issues/101.

oleq:
Closes #17.


src/editorui/editorui.js
```diff
@@ -51,6 +53,36 @@ export default class EditorUI extends Controller {
...
 		this.collections.add( new ControllerCollection( 'body' ) );
 	}
+
+	/**
+	 * Initializes EditorUI instance.
+	 *
+	 * @returns {Promise}
+	 */
+	init() {
+		this._setupIconManager();
```
Reinmar: Can't it be done in the constructor?

src/editorui/editorui.js
```diff
@@ -51,6 +53,36 @@ export default class EditorUI extends Controller {
...
+
+		return super.init();
+	}
+
+	/**
+	 * Adds IconManager into DOM.
+	 *
+	 * @protected
+	 */
+	_setupIconManager() {
```
Reinmar: Why a separate method?

src/editorui/editoruiview.js
```diff
@@ -44,7 +44,7 @@ export default class EditorUIView extends View {
 
 		this.applyTemplateToElement( bodyElement, {
 			attributes: {
-				class: 'ck-body'
+				class: 'ck-body ck-reset-all'
```
Reinmar: I don't think it's a good idea. Check what happened in http://tests.ckeditor.dev:1030/build/amd/tests/ckeditor5/creator/manual/creator-multi with the editor content.

src/iconmanagerview.js
```diff
@@ -0,0 +1,49 @@
...
+ * @license Copyright (c) 2003-2016, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+'use strict';
+
+import View from './view.js';
+
+/**
+ * Base class for the editor main views.
```
Reinmar: Sounds like copy-paste :D.

src/editorui/editorui.js
Reinmar: The same with `_setupSprite` below – I have a mixed feeling whether it makes sense to create separate methods for so simple things. It makes code harder to read instead of easier. I've seen it myself in my code – when I removed such methods the code became shorter and easier to read.

oleq: It's easier to inherit such class and override some specific methods than to override whole `init()` or `constructor()`. I don't believe anyone will ever refactor them when as too much code gets in there anyway, so we'd probably end up with a bloated `init()`/`constructor()` and nothing else. It's a pain for OOP approach. Besides, a specific method tells exactly what the class is supposed to do.

src/editorui/editoruiview.js
oleq: Probably not a good idea :+1:

src/editorui/editorui.js
Reinmar: > It's easier to inherit such class and override some specific methods than to override whole init() or constructor().

That would be a [square-rectangle problem](https://en.wikipedia.org/wiki/Circle-ellipse_problem). If you inherit, you should inherit, not override. So if we expect that anyone will want to replace icon manager, we need to find a different way to allow that – e.g. define `EditorUI` interface or base class.

Anyway, not an important issue here. Just wanted to point out that excessive code splitting may not make sense.

src/editorui/editoruiview.js
oleq: I think you meant that https://github.com/ckeditor/ckeditor5/pull/132/files#diff-2ea1cb288103b43043128e966e0050b7R32 instead.

Issue #3, opened 2016-03-15
View: An utility to extend existing templates


Reinmar:
Moved from https://github.com/ckeditor/ckeditor5-core/issues/219

oleq:
TL;DR: I decided to propose some general changes to Template–View interaction. 

First of all, I think the decision we made that event listener declaration should be as brief as possible like
```
{
  click: 'clicked'
}
```
is oversimplification and mixes Template with View logic. I think it's dirty. It silently says that `clicked` will be fired on `this` (`View` instance), which is not always true because it could be a `Model` or other `EmitterMixin` instance instead.

The same thing happened to attribute bindings. Using `View#attributeBinder.to()` implicates direct binding between the Template instance and `View#model`. It should be up to the Template, which `ObservableMixin` "class" instance it listens to and what it does to the DOM.

Then, once separated `View` from `Template` and vice–versa, the extension of templates can be done easily via `Template#extend`. The basic, class inheritance-based template extension could look like this:

```js
// IMO importing Template class in each View is not a huge cost.
import Template from './template.js';

class ButtonView extends View {
	constructor( model ) {
		super( model );

		this.template = new Template( {
			tag: 'button',

			attributes: {
				class: [
					'ck-button',
					// Move binding to Templates. It should be View-independent.
					// It should allow binding to any model, not View#model only.
					Template.bind.to( this.model, 'isEnabled', () => ... ) ),
					Template.bind.if( any model, 'isOn', () => ...  )

					// Alt syntax?
					//		[ 'to', this.model, 'isEnabled', () => { ... } ]
					//		[ 'if', any model, 'isOn', () => { ... } ]
				]
			},

			on: {
				click: [
					// Dispatching works for this View...
					Template.dispatch( this, 'clicked' ),

					// ...but also for any instance of EmitterMixin.
					Template.dispatch( any emitter, 'execute' ),

					() => { ... }

					// Alt syntax?
					// [ any emitter, 'clicked' ],
				]
			}
		} );

		// I think this method name is confusing. Why not
		//		this.region( 'children', el => el );
		// or
		//		this.setRegion( 'children', el => el );
		// or
		//		this.registerRegion( 'children', el => el );
		// ?
		this.register( 'children', el => el );
	}
}

class MyButtonView extends ButtonView {
	constructor( model ) {
		super( model );

		// Extending works via existing Template instance created by the parent class.
		this.template.extend( {
			attributes: {
				class: [
[...]

PR #21, opened 2016-04-06
t/20: Allow creating SVG elements via Template class

### LINKED ISSUE #20: Allow creating SVG elements via Template class
At the moment it is impossible because it requires `document.createElementNS` and `el.[set|get|remove]AttributeNS` API. SVG elements created with `document.createElement` won't work, i.e. they don't render icons.

oleq:
Closes #20.

### DIFF
--- a/src/editableui/editableuiview.js
+++ b/src/editableui/editableuiview.js
@@ -48,8 +48,8 @@ export default class EditableUIView extends View {
 
 		this.applyTemplateToElement( editableElement, {
 			attributes: {
-				contentEditable: bind.to( 'isEditable' ),
-				class: [ bind.to( 'isFocused', value => value ? 'ck-focused' : 'ck-blurred' ) ]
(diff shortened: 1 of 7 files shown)


src/editableui/editableuiview.js
```diff
@@ -48,8 +48,8 @@ export default class EditableUIView extends View {
 
 		this.applyTemplateToElement( editableElement, {
 			attributes: {
-				contentEditable: bind.to( 'isEditable' ),
-				class: [ bind.to( 'isFocused', value => value ? 'ck-focused' : 'ck-blurred' ) ]
+				contenteditable: bind.to( 'isEditable' ),
+				class: bind.to( 'isFocused', value => value ? 'ck-focused' : 'ck-blurred' )
```
Reinmar: Please see https://github.com/ckeditor/ckeditor5-ui/issues/3. That array was there for purpose.

PR #24, opened 2016-05-09
T/22: Provide necessary API for Classic Creator to happen

### LINKED ISSUE #22: Provide necessary tools for the first creator to emerge
The idea described in https://github.com/ckeditor/ckeditor5/issues/135.
The actual creator in https://github.com/ckeditor/ckeditor5-creator-classic/issues/1.

oleq:
As in https://github.com/ckeditor/ckeditor5/issues/135.

Closes #22.

### DIFF
--- a/src/boxededitorui/boxededitorui.js
+++ b/src/boxededitorui/boxededitorui.js
@@ -1,0 +1,38 @@
+/**
+ * @license Copyright (c) 2003-2016, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+'use strict';
+
+import EditorUI from '/ckeditor5/ui/editorui/editorui.js';
+import ControllerCollection from '/ckeditor5/ui/controllercollection.js';
+
+export default class BoxedEditorUI extends EditorUI {
+	constructor( editor ) {
+		super( editor );
+
+		this.collections.add( new ControllerCollection( 'top' ) );
+		this.collections.add( new ControllerCollection( 'editable' ) );
+
+		const config = editor.config;
+
+		/**
+		 * @property {Number} width
+		 */
+		this.set( 'width', config.get( 'ui.width' ) );
+
+		/**
+		 * @property {Number} height
+		 */
+		this.set( 'height', config.get( 'ui.height' ) );
+	}
+
+	/**
+	 * @readonly
+	 * @property {Model} viewModel
+	 */
+	get viewModel() {
+		return this;
+	}
+}
--- a/src/boxededitorui/boxededitorui.js
+++ b/src/boxededitorui/boxededitorui.js
@@ -18,11 +18,15 @@ export default class BoxedEditorUI extends EditorUI {
 		const config = editor.config;
 
 		/**
+		 * TODO: Handle editor resizing.
+		 *
 		 * @property {Number} width
 		 */
 		this.set( 'width', config.get( 'ui.width' ) );
 
 		/**
+		 * TODO: Handle editor resizing.
+		 *
 		 * @property {Number} height
 		 */
 		this.set( 'height', config.get( 'ui.height' ) );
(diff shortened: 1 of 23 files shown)


oleq:
Possible (obligatory?) TODO: `BoxedEditorUIView` resizing controlled by `BoxedEditorUI#[width,height]` attributes.

src/domemittermixin.js
```diff
@@ -268,7 +268,7 @@ function getNodeUID( node ) {
 // @param {Node} node
 // @return {Boolean} True when native DOM Node.
 function isDomNode( node ) {
-	return node && node.nodeType && ( typeof node.addEventListener == 'function' );
+	return node && ( typeof node.addEventListener == 'function' );
```
Reinmar: While this check should be ok in the future, because I suppose we're not going to create `addEventListener` methods, I would use `_.isNative`, just to not think about about it.

src/editableui/editableuiview.js
```diff
@@ -34,9 +34,12 @@ export default class EditableUIView extends View {
 	 * Sets the {@link #editableElement} property and applies necessary bindings to it.
 	 *
 	 * @param {HTMLElement} editableElement
+	 * @param {ui.TemplateDefinition} def
 	 */
-	setEditableElement( editableElement ) {
+	setEditableElement( editableElement, def = { attributes: {} } ) {
```
Reinmar: OK... this is ugly. And the way how this method must merge those templates. We've got to find a better solution. Can be a followup of this ticket if it requires some big issue in the view or somewhere.

Reinmar: I mean... the param may be acceptable, but it's dirty anyway and will be confusing for users. It's more important that we find some way to extend existing templates to simplify this method internally. But I'd love to get rid of this param too. This is an important part of this API unfortunately.

oleq:
WDYT about resize feature for `BoxedEditorUI`? A followup or here and now?

Reinmar:
WDYM by resize feature? A resize handle or a programatic way? There was a programatic way by changing `editorUI#width|height`

oleq:
Programatic way. It was for old `creator-classic` test (now `creator-legacy`). It's not implemented in `ClassicCreator` ATM though. Hence the question.

Reinmar:
Let's report a followup for that.

oleq:
Reported #25.

src/editableui/editableui.js
```diff
@@ -38,5 +38,6 @@ export default class EditableUI extends Controller {
 		 */
 		this.viewModel = new Model();
 		this.viewModel.bind( 'isEditable', 'isFocused' ).to( editableModel );
+		this.viewModel.set( 'editableName', editableModel.name );
```
Reinmar: Missing docs. In fact, the whole view model isn't documented.

src/boxededitorui/boxededitoruiview.js
```diff
@@ -0,0 +1,61 @@
...
+ * @license Copyright (c) 2003-2016, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+'use strict';
+
+import EditorUIView from '/ckeditor5/ui/editorui/editoruiview.js';
+import uid from '/ckeditor5/utils/uid.js';
+
+export default class BoxedEditorUIView extends EditorUIView {
```
Reinmar: Missing docs.

src/boxededitorui/boxededitorui.js
```diff
@@ -0,0 +1,42 @@
...
+ * @license Copyright (c) 2003-2016, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+'use strict';
+
+import EditorUI from '/ckeditor5/ui/editorui/editorui.js';
+import ControllerCollection from '/ckeditor5/ui/controllercollection.js';
+
+export default class BoxedEditorUI extends EditorUI {
```
Reinmar: Missing docs.

PR #39, opened 2016-06-02
T/3: An utility to extend existing templates

### LINKED ISSUE #3: View: An utility to extend existing templates

oleq:
Closes #3. 

Note: It requires changes in each View of the project to align template definitions to the new API first.


tests/template.js
```diff
@@ -10,21 +10,78 @@
...
+		it( 'accepts template definition', () => {
 			const def = {
 				tag: 'p'
 			};
 
 			expect( new Template( def ).definition ).to.equal( def );
 		} );
+
+		it( 'normalizes template definition', () => {
+			const bind = Template.bind( new Model( {} ), extend( {}, DOMEmitterMixin ) );
```
Reinmar: You can use `Object.create()` instead of `extend( {} )`.

Reinmar:
I can see that extension of children work as concatenation. Are you sure that this makes the most sense? How can you now extend existing children with new props? Would that work?

```js
const tpl1 = new Template( {
    tag: 'p',
    children: [
        {
            tag: 'a'
        }
    ]
} );

tpl1.children[ 0 ].extend( { attributes: { foo: 1 } );
```

If no, then there's no way to extend children...

oleq:
Nope. Extension of children works [as following](https://github.com/ckeditor/ckeditor5-ui/blob/t/3/tests/template.js#L1464-L1541): if the child index match in base and extension, it is extended recursively. If not (extension contains more children that base), a new child is added.

Reinmar:
Hm... Don't you think that this will be extremely surprising? And very prone to bugs. Someone changes number of children in the parent template (let's say this happens in the core UI lib) and then your extensions suddenly replace the children instead of extending them.

Reinmar:
Since it's super easy to just add or remove children from the template by modifying the array, I'm for recursive extensions as the only behaviour.

The only problem would be if a parent template was already rendered and you change the array, but I guess the current implementation wouldn't override the children as well? Is it actually possible to extend a template which was already rendered? If no, then I'm all for recursive extensions, because you can do other modifications in a different way.

oleq:
BTW:

```
tpl1.children[ 0 ].extend( { attributes: { foo: 1 } );
```

`tpl1.children[ 0 ]` is not an instance of `Template`.

oleq:
> The only problem would be if a parent template was already rendered and you change the array, but I guess the current implementation wouldn't override the children as well? Is it actually possible to extend a template which was already rendered? If no, then I'm all for recursive extensions, because you can do other modifications in a different way.

Once rendered, any modification to `Template` definition does not change anything in DOM. You got to re-render the template to obtain an element with your changes.

BTW: I'm pretty confused by "recursive extensions" terms you use. If we don't want any confusion like

> Hm... Don't you think that this will be extremely surprising? And very prone to bugs. Someone changes number of children in a parent template (let's say this happens in the core UI lib) and then your extensions suddenly replace the children instead of extending them.

the only solution is to make each child an instance of `Template` and `extend()` it the following way:

```
tpl1.children[ 0 ].extend( { attributes: { foo: 1 } );
```

which also means that any `children: [ ... ]` definition inside `tpl.extend()` is ignored. And it means lots of `Template` instances.

...but if you think about it, it doesn't solve the problem really because when

> Someone changes number of children in a parent template

then `tpl1.children[ INDEX ]` could also point to the wrong child and fall in the same trap as `tpl.extend()` with `children` definition.

oleq:
```js
// Parent class
	tag: 'p',
	children: [
		{
			tag: 'span',
			attributes: {
				class: 'foo'
			}
		}
	]
} );

// -> <p><span class="foo"></span></p>

// Class which inherits parent
this.template.extend( {
	children: [
		{
			attributes: {
				class: 'bar'
			}
		}
	]
} );

// -> <p><span class="foo bar"></span></p>

// Parent class gets upgraded to...
	tag: 'p',
	children: [
		{
			tag: 'b',
			children: [ 'Surprise' ]
		},
		{
			tag: 'span',
			attributes: {
				class: 'foo'
			}
		}
	]
} );

// so the class which inherits extends <b> instead of <span>...
this.template.extend( {
	children: [
		{
			attributes: {
				class: 'bar'
			}
		}
	]
} );

// -> <p><b class="bar">Surprise</b><span class="foo"></span></p>

// -------------------------------------------------------------------

// But the same would happen if extension was index-based...
this.template.children[ 0 ].extend( {
	attributes: {
		class: 'bar'
	}
} );

// because `this.template.children[ 0 ]` points to <b> now.
```

Reinmar:
This looks bad. I mean – every option that we mentioned so far sounds bad unfortunately because every single one will fail in some pretty basic scenario.

So I've got another idea – `extend()` should do a flat (1st level only) extension.

And if you want to extend children just do:

```js
Template.extend( tpl1.children[ nowGuessTheIndex ], {
    // child extension
} );
```

In this situation the only risky point is tell the index, but that's something you won't be able to simplify. But at least it's super clear what's happening here.

PS. This also means that `extend()` may need to be moved to the `Template`.

Reinmar:
Alternative to the option I proposed above is a `tpl.extend()` working recursively, but throwing when numbers of children don't match.

oleq:
> Alternative to the option I proposed above is a tpl.extend() working recursively, but throwing when numbers of children don't match.

It's not an alternative because it will disallow extension of children in such cases. I mean, there will be **no** other way to do this.

oleq:
It isn't also an alternative because it will fail if the base class children change order but the total number remains the same.

oleq:
> Alternative to the option I proposed above is a tpl.extend() working recursively, but throwing when numbers of children don't match.

On the other hand... I like the idea. If they want to add new children, they can use `this.template.definition.children.push( { ... } )`. So how about that?

Reinmar:
> On the other hand... I like the idea. If they want to add new children, they can use this.template.definition.children.push( { ... } ). So how about that?

That's exactly the idea. You can always modify children of an existing template by simply working with them as with an array. You can retrieve one and modify using `Template.extend()`, you can remove one, add one, etc.

That's why I think that `Template.extend()` should work only one one level.

PS. note that this means that two types can be passed to `extend()` as its first param:  `Template.extend( {Template|TemplateDefinition}, TemplateDefinitionFragment );`

oleq:
> PS. note that this means that two types can be passed to extend() as its first param: Template.extend( {Template|TemplateDefinition}, TemplateDefinitionFragment );

It does not matter. The top-level definition is no different that any element of its `children: [ ... ]`.

Reinmar:
I only meant that it may be a `Template` instance OR `TemplateDefintion`. I meant the first param.

Reinmar:
I pushed small commit – please check whether I defined the param types in API docs correctly.

oleq:
Please note that this PR and related sub-PRs move event fireing logic from `View` to `View#model`. It's a foretaste of #26.

src/editableui/editableuiview.js
```diff
@@ -23,22 +24,20 @@ export default class EditableUIView extends View {
...
 			this.element = this.editableElement = editableElement;
 		}
 
-		this.template = {
+		this.template = new Template( {
 			tag: 'div',
 			attributes: {
 				class: [
-					bind.to( 'isFocused', value => value ? 'ck-focused' : 'ck-blurred' ),
+					this.bind.to( 'isFocused', value => value ? 'ck-focused' : 'ck-blurred' ),
```
Reinmar: I'd really consider always defining `bind` const at the beginning of a constructor and then using it without `this`. Just like we do with the `t()` function. It will better show that this is not View's stuff. WDYT?

src/editableui/editableuiview.js
```diff
@@ -54,7 +53,7 @@ export default class EditableUIView extends View {
 	 */
 	init() {
 		if ( this.editableElement ) {
-			this.applyTemplateToElement( this.editableElement, this.template );
+			this.applyTemplateToElement( this.editableElement, this.template.definition );
```
Reinmar: Can't you do this through `this.template.apply( this.editableElement )`?

src/editorui/boxed/boxededitoruiview.js
```diff
@@ -7,6 +7,7 @@
 
 import EditorUIView from '../../editorui/editoruiview.js';
 import uid from '../../../utils/uid.js';
+import Template from '/ckeditor5/ui/template.js';
```
Reinmar: That's a thing I haven't thought initially... additional import every time you do something with a template. Maybe it makes sense to publish this class as view's property (static or instance) as well? In the end, view will pretty much never be used without the template... Unless you want to use a different templating system with our mvvm architecture, but it's an edge case.

oleq: Yeah, this is something I was thinking about after I published this PR. Since we switched from `tplInstance.extend()` to `Template.extend()` we could go back to `this.template = ui.TemplateDefinition` in View instead of `this.template = ui.Template` to avoid that unnecessary import. All the internals would be in the background again.

src/editableui/editableuiview.js
oleq: Sounds good. But when there's a single binding only in the definition, an access through `this.bind` is enough IMO to keep the code clutter-free.

src/editableui/editableuiview.js
oleq: If we stay with the concept that `this.template` in an instance of `Template` – I guess so. Otherwise – not really. Or maybe with `this._template.apply( ... )`, which is not very cool anyway.

src/editorui/boxed/boxededitoruiview.js
oleq: On the other hand it makes `this.template.apply( ... )` [impossible](https://github.com/ckeditor/ckeditor5-ui/pull/39#discussion_r65671246).

src/editableui/editableuiview.js
```diff
@@ -6,6 +6,7 @@
 'use strict';
 
 import View from '../view.js';
+import Template from '/ckeditor5/ui/template.js';
```
Reinmar: Don't use abs paths.

src/template.js
```diff
@@ -21,6 +28,8 @@ export default class Template {
 	 * @param {ui.TemplateDefinition} def The definition of the template.
 	 */
 	constructor( def ) {
+		normalize( def );
```
Reinmar: Normalization should not modify the passed object.

Reinmar:
Not sure what has changed but I don't have 100% CC in List controller:

```js
			for (let itemModel of this.model.items) {
				this._addListItem(itemModel);
			}
```

_addListItem isn't called even once. Please fix this in this PR so we have a clear picture.

src/template.js
```diff
@@ -305,10 +633,277 @@ function getElementAttributeUpdater( el, attrName, ns = null ) {
...
+			obj[ a ] = ext[ a ];
+		}
+	}
+}
+
+/**
+ * A helper for {@link ui.Template#extend}. Recursively extends {@link ui.Template#definition}
+ * with content from another definition. See {@link ui.Template#extend} to learn more.
+ *
+ * @ignore
```
Reinmar: Instead of using `@private` use `//` comments for stuff that should be ignored.

src/template.js
```diff
@@ -238,40 +258,345 @@ export default class Template {
...
+		 *		} ).render();
+		 *
+		 * @static
+		 * @method ui.Template.bind#if
+		 * @param {String} attribute An attribute name of {@link utils.ObservableMixin} used in the binding.
+		 * @param {String} [valueIfTrue] Value set when {@link utils.ObservableMixin} attribute is not undefined/null/false/''.
+		 * @param {Function} [callback] Allows processing of the value. Accepts `Node` and `value` as arguments.
+		 * @return {ui.TemplateBinding}
+		 */
+		if: ( attribute, valueIfTrue, callback ) => {
```
Reinmar: It can be `if( ... ) {`.

src/template.js
```diff
@@ -238,40 +258,345 @@ export default class Template {
...
+		 * @static
+		 * @method ui.Template.bind#to
+		 * @param {String} attribute Name of {@link utils.ObservableMixin} used in the binding.
+		 * @param {Function} [callback] Allows processing of the value. Accepts `Node` and `value` as arguments.
+		 * @return {ui.TemplateBinding}
+		 */
+		to: ( eventNameOrFuncionOrAttribute, callback ) => {
+			return {
+				type: bindToSymbol,
+				eventNameOrFuncion: eventNameOrFuncionOrAttribute,
```
Reinmar: Interesting typo all over the code :D.

src/template.js
```diff
@@ -238,40 +258,345 @@ export default class Template {
...
+ * An entry point to the interface which allows binding DOM nodes to {@link utils.ObservableMixin}.
+ * There are two types of bindings:
+ *
+ * * `HTMLElement` attributes or Text Node `textContent` can be synchronized with {@link utils.ObservableMixin}
+ * instance attributes. See {@link ui.Template.bind#to} and {@link ui.Template.bind#if}.
+ *
+ * * DOM events fired on `HTMLElement` can be propagated through {@link utils.ObservableMixin}.
+ * See {@link ui.Template.bind#to}.
+ *
+ * @property ui.Template#bind
```
Reinmar: It's a static method, not a property.

src/template.js
```diff
@@ -238,40 +258,345 @@ export default class Template {
 	 *
 	 * @protected
 	 * @param {ui.TemplateDefinition} def Definition of an element.
-	 * @param {HTMLElement} el Element which is rendered.
+	 * @param {HTMLElement} el Element which is being rendered.
 	 */
-	_activateElementListenerAttachers( def, el ) {
+	_setupListeners( def, el ) {
```
Reinmar: `s/_setUpListeners/`

src/template.js
```diff
@@ -238,40 +258,345 @@ export default class Template {
...
-			.map( name => [ name, ...name.split( '@' ) ] )
-			.forEach( split => {
-				// TODO: ES6 destructuring.
-				const key = split[ 0 ];
-				const evtName = split[ 1 ];
-				const evtSelector = split[ 2 ] || null;
+		for ( let key in def.on ) {
+			const [ domEvtName, domSelector ] = key.split( '@' );
+
+			def.on[ key ].forEach( schemaItem => {
```
Reinmar: If you want @pjasiun to love you you can replace `forEach` with for-of loop.

src/template.js
```diff
@@ -95,7 +120,7 @@ export default class Template {
 			throw new CKEditorError( 'ui-template-wrong-syntax' );
```
Reinmar: Misses human-readable message after `:`.

src/template.js
```diff
@@ -8,10 +8,17 @@
...
 
 import CKEditorError from '../utils/ckeditorerror.js';
+import mix from '../utils/mix.js';
+import EmitterMixin from '/ckeditor5/utils/emittermixin.js';
+
+const bindToSymbol = Symbol( 'bindTo' );
+const bindIfSymbol = Symbol( 'bindIf' );
 
 /**
  * Basic Template class.
```
Reinmar: You could make it a bit nicer :D.

Reinmar:
I left couple comments in the code + there's been an old one about using `bind` without `this`.

Reinmar:
All merged. Thanks, @oleq!

Issue #1, opened 2016-03-15
Move toolbar.addButtons to its model and create model->view binding for buttons


Reinmar:
Moved from https://github.com/ckeditor/ckeditor5-core/issues/235.

Reinmar:
Strongly related to #34.

PR #41, opened 2016-06-27
T/26: Review our approach to MVVM

### LINKED ISSUE #26: Review our approach to MVVM
Currently, we've implemented MVC or MV(C)VM, depending on a component. There's been ongoing discussion between us on which approach is better and we couldn't make a final decision, so the code isn't very consistent at the moment.

We need to verify which pattern better fit into our lib. So far we think that MV(C)VM will do best, however, we need to improve the initialisation of such components.

Currently, there's an issue with how view model can be created by a controller (and then passed to view), while controller should also be given its view in its constructor. This means initialisation like this:

```js
const model = new Model();
const controller = new Controller( model );
const view = new View( controller.viewModel );

controller.view = view;
```

This is obviously ugly. Fortunately, we seem to agree that since view model is very tightly bound to the view, it can be instantiated by that view, to be later retrieved and controlled by the controller:

```js
const model = new Model();
const view = new View(); // it creates view.model
const controller = new Controller( model, view ); // controller reads view.model
```

oleq:
Closes #26.


Reinmar:
See https://github.com/ckeditor/ckeditor5-editor-classic/pull/15#discussion_r68746767

src/bindings/toolbarbindingsmixin.js
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+'use strict';
+
+/**
+ * Mixin that injects the common Toolbar–like bindings.
+ *
+ * @mixin ui.bindings.ToolbarBindingsMixin
+ */
+const ToolbarBindingsMixin = {
```
Reinmar: Why have you extracted this from the toolbar? The whole `addButtons()` needs to be fixed (see https://github.com/ckeditor/ckeditor5-ui/issues/1) in general, so if such a change was required, perhaps we should also fix that issue at once.

src/bindings/stickytoolbar.js
```diff
@@ -0,0 +1,34 @@
...
+	 * @param {ckeditor5.Editor} editor
+	 */
+	constructor( model, view, editor ) {
+		super( model, view );
+
+		this.editor = editor;
+	}
+}
+
+Object.assign( StickyToolbar.prototype, ToolbarBindingsMixin );
```
Reinmar: I don't think we need that mixin anyway, but if we do, we have a special `mix()` function in utils.

src/bindings/toolbarbindingsmixin.js
oleq: I extracted it because bindings/toolbar and bindings/stickytoolbar use the same `addButtons` method. Please don't mix the issues, this one is huge enough. If we decide to remove it, it will be done in #1. I can handle it immediately after #26.

Reinmar:
Great job! IMO, the components are now a lot, lot cleaner. I've only found couple of small things (mainly docs) which I'll report in a followup.

Reinmar:
Turns out there's a failing test as well – https://github.com/ckeditor/ckeditor5-ui-default/issues/25. I haven't caught it because I haven't checked that one of the branches wasn't rebased on latest master.

Reinmar:
Other issues:

* https://github.com/ckeditor/ckeditor5-ui-default/issues/26
* https://github.com/ckeditor/ckeditor5-ui/issues/43
* https://github.com/ckeditor/ckeditor5-ui/issues/44

PR #42, opened 2016-06-28
Support for binding styles to view model props

### LINKED ISSUE #27: Support for binding styles to view model props
So far we needed to something pretty ugly: https://github.com/ckeditor/ckeditor5/pull/152/files#diff-8141b614b20646bb5ecb4cd49dcb19eaR32

That needs to be improved.

oskarwrobel:
Fixes #27

### DIFF
--- a/src/template.js
+++ b/src/template.js
@@ -23,7 +23,10 @@ const bindIfSymbol = Symbol( 'bindIf' );
  *		new Template( {
  *			tag: 'p',
  *			attributes: {
- *				class: 'foo'
+ *				class: 'foo',
+ *				style: {
+ *					backgroundColor: 'yellow'
+ *				}
  *			},
  *			children: [
  *				'A paragraph.'
@@ -32,7 +35,7 @@ const bindIfSymbol = Symbol( 'bindIf' );
  *
  * will render the following HTMLElement:
  *
- *		<p class="foo">A paragraph.</p>
+ *		<p class="foo" style="backgroundColor:yellow">A paragraph.</p>
  *
  * See {@link ui.TemplateDefinition} to know more about templates and see complex examples.
  *
@@ -402,6 +405,14 @@ export default class Template {
 			attrValue = attributes[ attrName ];
 			attrNs = attrValue[ 0 ].ns || null;
 
+			// Attribute style has specific format so needs to be parsed in a specific way:
+			// 		input: { style: { property: value, otherProperty: Template.bind.to( ... ) } }
+			// 		output: style="property:value;other-property:otherValue"
+			if ( attrName == 'style' ) {
+				this._renderAttributeStyle( attrValue[ 0 ].value || attrValue[ 0 ], attrNs, el );
+				continue;
+			}
+
 			// Activate binding if one is found. Cases:
 			// 		{ class: [ Template.bind.to( ... ) ] }
 			// 		{ class: [ 'bar', Template.bind.to( ... ), 'baz' ] }
@@ -432,6 +443,60 @@ export default class Template {
 				el.setAttributeNS( attrNs, attrName, attrValue );
 			}
 		}
+	}
+
+	/**
+	 * Render attribute `style`.
+	 * Attribute style has specific format so needs to be parsed in a specific way. For example:
+	 *
+	 * 		style: {
+	 * 			property: value,
+	 * 			otherProperty: otherValue
+	 * 		}
+	 *
+	 * 	will be rendered as:
+	 *
+	 * 		style="property:value;other-property:otherValue"
+	 *
+	 * Note: Multiple-word properties are always defined in camelCase format, and then are parsed to dash format.
+	 *
+	 * @private
+	 * @param {ui.TemplateDefinition.attributes.styles} styles Styles definition. Multiple-word properties needs to be
+	 * defined in camelCase format.
+	 * @param {String} attrNs Attribute namespace.
+	 * @param {HTMLElement} el Element which is rendered.
+	 */
+	_renderAttributeStyle( styles, attrNs, el ) {
+		// Normalization in case of style attribute is created with additional data as custom namespace.
+		styles = styles[ 0 ] || styles;
+
+		// Iterate through every single style.
+		// Attach listener to style property if is observable and get initial value for each style.
+		const initialStyles = Object.keys( styles ).map( ( style ) => {
+			// If style value is not observable.
+			if ( !styles[ style ].observable ) {
+				// Just return style as `property-name: value`.
+				// Note that style property has to be transformed from camelCase to dash
+				// for setting initial value by `setAttributeNS()`.
+				return `${ camelCaseToDash( style ) }:${ styles[ style ] }`;
+			}
+
+			// If style value is observable.
+			let { emitter, observable, attribute } = styles[ style ];
+
+			// Add listener.
+			emitter.listenTo( observable, `change:${ attribute }`, ( eventInfo, property, value ) => {
+				// And update only this style which has been changed.
+				el.style[ style ] = value;
+			} );
+
+			// A the end return style as `property-name: value`.
+			// Note that style property has to be transformed from camelCase to dash
+			// for setting initial value by `setAttributeNS()`.
+			return `${ camelCaseToDash( style ) }:${ styles[ style ].observable[ attribute ] }`;
+		} );
+
+		el.setAttributeNS( attrNs, 'style', initialStyles.join( ';' ) );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


src/template.js
```diff
@@ -435,6 +444,109 @@ export default class Template {
 	}
 
 	/**
+	 * Render attribute `style`.
+	 *
+	 * Attribute style value could be an {Object} with static or binded to model properties:
```
oleq: s/binded/bound

src/template.js
```diff
@@ -435,6 +444,109 @@ export default class Template {
...
+	 *		} );
+	 *
+	 *		attributes: {
+	 * 			style: {
+	 * 				property: value,
+	 * 				otherProperty: bind.to( 'modelProperty' ),
+	 * 			}
+	 * 		}
+	 *
+	 * 	or a {String} if whole style attribute is binded to model property:
```
oleq: s/binded/bound

src/template.js
```diff
@@ -435,6 +444,109 @@ export default class Template {
...
+	 * 1. It seems to be not necessary
+	 * 2. We are using more efficient methods for setting styles `el.style.property` and `el.style.cssText` instead of
+	 * `setAttributeNS()`
+	 *
+	 * @private
+	 * @param {ui.TemplateDefinition.attributes.styles} styles Styles definition.
+	 * @param {HTMLElement} el Element which is rendered.
+	 */
+	_renderAttributeStyle( styles, el ) {
+		// If whole attribute is binded we don't parse value.
```
oleq: s/binded/bound

tests/template.js
```diff
@@ -315,6 +315,102 @@ describe( 'Template', () => {
...
+
+				observable.width = '50px';
+				observable.color = 'green';
+
+				expect( el.style.width ).to.be.equal( '50px' );
+				expect( el.style.height ).to.be.equal( '200px' );
+				expect( el.style.backgroundColor ).to.be.equal( 'green' );
+			} );
+
+			it( 'renders fully binded', () => {
```
oleq: s/binded/bound

tests/template.js
```diff
@@ -315,6 +315,102 @@ describe( 'Template', () => {
...
+
+				expect( el.style.width ).to.be.equal( '100px' );
+
+				observable.style = 'width:20px;height:50px';
+
+				expect( el.style.width ).to.be.equal( '20px' );
+				expect( el.style.height ).to.be.equal( '50px' );
+			} );
+
+			it( 'throws when wrong format for initial value of fully binded', () => {
```
oleq: s/binded/bound

tests/template.js
```diff
@@ -315,6 +315,102 @@ describe( 'Template', () => {
...
+					new Template( {
+						tag: 'p',
+						attributes: {
+							style: bind.to( 'style' )
+						}
+					} ).render();
+				} ).to.throw( CKEditorError, /template-renderAttributeStyle-invalid-format/ );
+			} );
+
+			it( 'throws when wrong format for updated value of fully binded', () => {
```
oleq: s/binded/bound

tests/template.js
```diff
@@ -315,6 +315,102 @@ describe( 'Template', () => {
...
+
+		describe( '`style` attribute', () => {
+			it( 'renders with dynamic and static properties', () => {
+				const observable = new Model( {
+					width: '100px',
+					color: 'yellow'
+				} );
+
+				const emitter = Object.create( EmitterMixin );
+				const bind = Template.bind( observable, emitter );
```
oleq: Use `beforeEach( ()=> {} )` to avoid duplication of emitter/binder creation in each test.

tests/template.js
```diff
@@ -315,6 +315,102 @@ describe( 'Template', () => {
...
+					attributes: {
+						style: {
+							width: bind.to( 'width' ),
+							height: '200px',
+							backgroundColor: bind.to( 'color' )
+						}
+					}
+				} ).render();
+
+				expect( el.style.width ).to.be.equal( '100px' );
```
oleq: s/to.be.equal/to.equal

src/template.js
```diff
@@ -402,6 +405,12 @@ export default class Template {
 			attrValue = attributes[ attrName ];
 			attrNs = attrValue[ 0 ].ns || null;
 
+			// Attribute style might have specific format so needs to be parsed in a specific way.
+			if ( attrName == 'style' ) {
+				this._renderAttributeStyle( attrValue[ 0 ].value || attrValue[ 0 ], el );
+				continue;
```
oleq: This `if` is followed by `if...else` statements. Why not replace the next one `if` to `else if`? I think `continue` should be avoided when possible. It's more bytes and out of local algorithm flow convention.

oleq:
I think that you duplicated a lot of code in your solution instead of re–using existing methods. I managed to achieve the same result (- error handling) with the following:

```js
_renderStyleAttribute( styles, el ) {
	// style: bind.to( 'attribute' )
	if ( hasBinding( styles ) ) {
		this._bindToObservable( styles, el, getAttributeUpdater( el, 'style' ) );
	}
	// style: { ... }
	else {
		for ( let styleName in styles[ 0 ] ) {
			const styleValue = styles[ 0 ][ styleName ];

			// style: { color: bind.to( 'attribute' ) }
			if ( hasBinding( styleValue ) ) {
				this._bindToObservable( [ styleValue ], el, getStyleUpdater( el, styleName ) );
			}
			// style: { color: 'red' }
			else {
				el.style[ styleName ] = styleValue;
			}
		}
	}
}
```

and 

```js
function getStyleUpdater( el, styleName ) {
	return {
		set( value ) {
			el.style[ styleName ] = value;
		},

		remove() {
			delete el.style[ styleName ];
		}
	};
}
```

src/template.js
```diff
@@ -435,6 +444,109 @@ export default class Template {
...
+	 * Note: Attribute `style` is rendered without setting namespace because:
+	 * 1. It seems to be not necessary
+	 * 2. We are using more efficient methods for setting styles `el.style.property` and `el.style.cssText` instead of
+	 * `setAttributeNS()`
+	 *
+	 * @private
+	 * @param {ui.TemplateDefinition.attributes.styles} styles Styles definition.
+	 * @param {HTMLElement} el Element which is rendered.
+	 */
+	_renderAttributeStyle( styles, el ) {
```
oleq: My common sense tells me that it sould be `_renderStyleAttribute` instead of `_renderAttributeStyle`. The second suggests that the attribute has style, which is not true, isn't it?

oleq:
I also see no test which would check what happens when Model attribute value is "falsy", like in [this](https://github.com/ckeditor/ckeditor5-ui/blob/t/27/tests/template.js#L1023) test.

oleq:
BTW: Why `el.style.cssText`? Is it faster or simpler? I didn't research this but for me `el.styles` object is the easiest way to access CSS styles. 

I guess that changing `el.style.cssText` when a single component (like `color`) of the `style="..."` attribute has changed would cause a re–paint in DOM because the browser could interpret it that all the components have changed. `el.styles.component` feels more... direct. But as I said – I didn't research the matter and I'm not quite sure to what extend browsers optimise such operations.

oskarwrobel:
Thanks for great review.

> Why el.style.cssText

It was for setting initial value only. To set every style at once instead of setting single style by `el.style[property] = value;` in a loop to avoid unnecessary re-paints.

oskarwrobel:
> I think that you duplicated a lot of code in your solution instead of re–using existing methods. I managed to achieve the same result (- error handling) with the following:

If we don't need validation and throwing error (I didn't have it in my first version) we can do it like this:
```javascript
if ( hasBinding( attrValue ) ) { // This will handle style: bind.to( 'style' )
    // ...
} else if ( attrName == 'style' ) {
    this._renderStyleAttribute( attrValue[ 0 ].value || attrValue[ 0 ], el );
} else {
    // ...
}

oleq:
> To set every style at once instead of setting single style by el.style[property] = value; in a loop to avoid unnecessary re-paints.

Again, I'm not sure how browsers optimise things. They could re–paint synchronously as `el.style.foo` changes or aggregate multiple changes and asynchronously re–paint in some intervals, as a batch. I guess we could check it in dev tools, could we? 

For the sake of keeping the code simple I would stick to iterating over `el.style` but it would be a nice to know how browsers work.

Reinmar:
Are you sure there will be repaints between style changes? I'd think that the browser will do everything in one batch unless you specifically trigger a repaint somehow. That's how it works for layout calculations.

oleq:
> If we don't need validation and throwing error

I think that less errors is better. We argued about it when implementing core UI classes. The conclusion was that in the places which will fail bloody red anyway, like when user uses wrong type of data, there is no need for a dedicated error. Errors should cover crucial issues, IMO.

oleq:
@Reinmar I guess so. But regardless of the facts, let's start with a simple solution first and then optimise it, if needed.

PR #49, opened 2016-07-25
t/34: Created API to sync list–like UI components with collections of models

### LINKED ISSUE #34: Create API to bind list–like UI components with contents of Collection–like classes
A followup of https://github.com/ckeditor/ckeditor5-ui-default/pull/15#discussion_r64541864.

At the moment `List` UI component implements own bindings between the collection in the model and collection of child controllers. It might be that some nicer, re–usable API is necessary to automate this process. More use–cases are needed to tell for sure though.

oleq:
Closes #34.


oleq:
Just FYI, there are still some minor things to be tweaked in this branch once we decide the idea is right.

PR #60, opened 2016-08-04
t/35: Implement event delegation mechanism

### LINKED ISSUE #35: Implement event delegation mechanism
Requires https://github.com/ckeditor/ckeditor5-utils/issues/50.

A followup of https://github.com/ckeditor/ckeditor5-ui-default/pull/15#discussion_r64554432.

At the moment there is no automated (configurable) event delegation mechanism between controllers and their children (also views and their children). This will be a common case in the future so manual delegation in listeners is a no go. 

Note: It may also be necessary for Models, like when a Collection or a Model is an attribute of another Model.

oleq:
…ils.ObservableMixin#pipe.

Closes #35.


scofalik:
In the issue you wrote that it requires changes in utils: https://github.com/ckeditor/ckeditor5-utils/issues/50 but it does not, also you have this in docs: https://github.com/ckeditor/ckeditor5-ui/pull/60/files#diff-9a387d89a32f80f40488cdabd2c657caR219 but `utils.ObservableMixin#pipe` is not used. It could be used like this:

```
// Activate piping on existing controllers in this collection.
for ( let controller of this ) {
	for ( let evtName of events ) {
		controller.model.pipe( evtName ).to( dest );
	}
}
```

To get rid of internal function and make the method a bit less complex. Was it your intend? Having this doubts I am not sure whether you pushed all the changes you did. Anyway, if we would like to use `pipe` method, it would be also nice to have some kind of `unpipe` method, to use here: https://github.com/ckeditor/ckeditor5-ui/pull/60/files#diff-9a387d89a32f80f40488cdabd2c657caR248
But it of course works as it is right now.

Also, adding and event on `add` and `remove` each time event is piped seems like a lot of events but this is probably a pre-mature optimization case :P.

BTW you use `ObservableMixin` a lot in docs, where it seems like `EmitterMixin` is enough. Or do I not see something?

oleq:
> Having this doubts I am not sure whether you pushed all the changes you did.

Yes. I pushed all I had at that time. Still it was more like "let's put it on review and see the feedback" than "it is merge–ready feature". Nice catch, anyway.

> Anyway, if we would like to use pipe method, it would be also nice to have some kind of unpipe method, to use here

As for piping in `ControllerCollection`, it is unlikely to happen that some collection will "unpipe". It would be an edge case or at least, there's no clear use–case ATM. It's more a one–way road when defining `constructor` of the `Controller` instance. Most of the definition lasts until the editor is destroyed. 

`EmitterMixin#unpipe` is another topic. I'll check this out.

> Also, adding and event on add and remove each time event is piped seems like a lot of events but this is probably a pre-mature optimization case :P.

Yes. It's not the first time I (we) decided to go with this kind of implementation (easy but slow) in UI because, in fact, it may turn out that that no optimisation will ever be needed. UI is seldom processing data and reacting to abrupt changes so for the sake of keeping code shorter and much clearer I think it's OK to sacrifice the performance and memory. Besides, as long as it is JS, not the DOM, it's rather a minor difference anyway.

> BTW you use ObservableMixin a lot in docs, where it seems like EmitterMixin is enough. Or do I not see something?

As you correctly pointed out in https://github.com/ckeditor/ckeditor5-utils/pull/56#issuecomment-237797863, the implementation will rather go to `EmitterMixin`. So all the things will move anyway.

PR #75, opened 2016-09-13
t/70: Template class refactoring for better child management 

### LINKED ISSUE #70: Template class needs refactoring to allow easier extension and child management
A followup of https://github.com/ckeditor/ckeditor5-ui-default/pull/60#issuecomment-245280359.

--- 

The last refac in the `Template` class defined `Template.extend` so that it allows extension of existing children but disallows adding new ones. The reasoning was that if some component is to have a different inner structure, it should not inherit but be created from scratch. Apparently, it was not a very good decision because in case of components like `BalloonPanel`, which is nothing but a container, sooner or later classes inheriting from it will require some extensions in the structure of the template.

Because of this decision it's very tricky to add new children to the template, i.e.

```js
Template.extend( this.template, {
    attributes: {
        class: [
            'ck-link-balloon-panel',
        ]
    }
} );

// Children must be added outside of Template.extend.
this.template.definition.children = [
    {
        tag: 'div',
        attributes: {
            // An attribute must be in "normalized" form because this.template
            // is an instance, which #definition has already been normalized.
            // So using "class: 'moo'"" would throw lots of errors at the moment of rendering.
            class: [
                'moo'
            ]
        }
    }
];
```        

So to simplify things, I propose another refactoring in the `Template` class:

1. ATM the structure is that an instance of `Template` has `#definition` (which is in a normalized form), which in turn has `#children`, which is an array of `#definition`(s). Because of that, no matter how the user wants to manipulate the children of the template, they must use and know how to write normalized definitions, which is a really nasty thing. The solution for it is: when constructor of the `Template` is called it normalizes its definition, so each member of `definition#children` becomes an **instance** of `Template`, if not already. This will allow the following:

    ```js
        tag: 'div',
        attributes: {
            class: [
                'ck-balloon-panel',
            ],
            children: [
                // This one is already a Template. Perfect!
                new Template( {
                    tag: 'div',
                    attributes: {
                        class: [
                            'some-child',
                        ],
                    }
                } ),
    
                // This one becomes an instance as soon as constructor is called.
                {
                    tag: 'div',
                    attributes: {
                        class: [
                            'another-child',
                        ],
                    }
                }
            ]
        }
    } );
    ```
    
    and the following
    
    ```js
    this.template.definition.children.push( new Template( { ... } ) );
    this.template.definition.children.splice( ... );
    ```

2. Get rid of `Template#definition`. As the `Template.constructor` accepts the definition object, its should create `Template.tag""`, `Template.attributes{}`, `Template.children[]` etc. It would simplify certain things, like

    ```js
    this.template.definition.children.push( new Template( { ... } ) );
    ```

    becomes
    
    ```js
    this.template.children.push( new Template( { ... } ) );
    ```

3. Update `Template` methods like `static extend` so they work with the new structure. Still it will not allow adding new children because it means a very, very tricky syntax, especially when there are already some children and the new one must land between them but **not** become and extension of one of them. What's more I'm for reducing the functionality of `Template.extend` so it cannot extend children because it will be easy enough to call `Template.extend( this.template.children[ 3 ], { ... } )` instead. So no more

   ```js
    Template.extend( instance, {
        attributes: {
            class: 'b',
        },
        children: [
            {
                attributes: {
                    class: 'c'
                }
            }
        ]
    } );
    ```

It might feel like a lot of work but it isn't, IMO. There are some tests which must be slightly altered or extended but the overall concept of the `Template` would remain. So it's not a complete re–write. This time :P

oleq:
Fixes #70.

### DIFF
--- a/src/template.js
+++ b/src/template.js
@@ -61,7 +61,7 @@ export default class Template {
 		/**
 		 * Text of this template, indicating that the instance will render to a DOM Text.
 		 *
-		 * @member {Array.<String,ui.TemplateValueSchema> ui.Template#text
+		 * @member {Array.<String|ui.TemplateValueSchema> ui.Template#text
 		 */
 
 		/**
@@ -312,7 +312,7 @@ export default class Template {
 	 * @protected
 	 * @param {Node} applyNode If specified, this template will be applied to an existing DOM Node.
 	 * @param {Boolean} intoFragment If set, children are rendered into `DocumentFragment`.
-	 * @returns {HTMLElement} A rendered Node.
+	 * @returns {HTMLElement|Text} A rendered Node.
 	 */
 	_renderNode( applyNode, intoFragment ) {
 		let isInvalid;
--- a/src/template.js
+++ b/src/template.js
@@ -8,14 +8,16 @@
 import CKEditorError from '../utils/ckeditorerror.js';
 import mix from '../utils/mix.js';
 import EmitterMixin from '../utils/emittermixin.js';
+import Collection from '../utils/collection.js';
 import cloneDeepWith from '../utils/lib/lodash/cloneDeepWith.js';
 import isObject from '../utils/lib/lodash/isObject.js';
 
 const bindToSymbol = Symbol( 'bindTo' );
 const bindIfSymbol = Symbol( 'bindIf' );
+const xhtmlNs = 'http://www.w3.org/1999/xhtml';
 
 /**
- * A basic Template class. It renders DOM HTMLElements from {@link ui.TemplateDefinition} and supports
+ * A basic Template class. It renders DOM HTMLElement or Text from {@link ui.TemplateDefinition} and supports
  * element attributes, children, bindings to {@link utils.ObservableMixin} instances and DOM events
  * propagation. For example:
  *
@@ -36,7 +38,7 @@ const bindIfSymbol = Symbol( 'bindIf' );
  *
  *		<p class="foo" style="background-color: yellow;">A paragraph.</p>
  *
- * See {@link ui.TemplateDefinition} to know more about templates and see complex examples.
+ * See {@link ui.TemplateDefinition} to know more about templates and complex template definitions.
  *
  * @memberOf ui
  */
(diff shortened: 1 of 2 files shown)


Reinmar:
LGTM, with one proposal. I'd consider splitting the `Template` class into templates for element and text nodes. Then we'd also need a `template()` function which gets a definition and returns a proper structure of `TemplateElement` and `TemplateText` classes.

oleq:
> I'd consider splitting the Template class into templates for element and text nodes.

Pros? Cons?

Reinmar:
> Pros? Cons?

Better because cleaner design. You have then classes which do one thing – render a specific type of element. E.g. now the template can have a tag name defined or text. So you need ugly validation which one of them is defined. And you repeat in multiple places that "this only works if a tag is defined". This is wrong.

With two classes, each rendering one type of thing, you'll avoid all these issues.

oleq:
> this only works if a tag is defined

It's merely a documentation. I think we should move on and consider it a followup because there are other issues which need attention and they, on the contrary, will bring some tangible value to MVP.

Reinmar:
> It's merely a documentation. I think we should move on and consider it a followup because there are other issues which need attention and they, on the contrary, will bring some tangible value to MVP.

It's not about a documentation. It's about how the code is organised. We can do this in a followup, but the current PR makes things a bit worse. At the same time, we don't plan to use the better extensions right now, so this ticket doesn't block anything today. If you want to split it into two, I'm fine, but the followup should be done asap anyway.

oleq:
> It's about how the code is organised.

It's been organised like this for months now. This class is a helper, unlikely to be extended in the nearest future. The change you're expecting is dedicated purely to satisfy your sense of aesthetics. It does not bring the value to the project, whether Template class is split in two or not.

> but the current PR makes things a bit worse

No, it does not. Because I enhanced some docs, which now highlight some aspects of the implementation, it didn't change the actual way the `Template` class works. Nothing has changed in that matter.

> we don't plan to use the better extensions right now

Any UI component we work with could possible use it any time. It's needed and the Link feature just  proved it.

> If you want to split it into two, I'm fine

And now you're talking like a human being 😄  

Besides, please note that the change you suggested is beyond the scope of this issue. It's not an "umbrella issue" to fix all the problems of this planet, but a specific fix to a specific problem. We often forget about it.

PR #82, opened 2016-09-15
Added event capturing option to DOMEmitterMixin.

### LINKED ISSUE #76: Lack of useCapture option in DOMEmitterMixin
There is no way to set `useCapture` option in `DOMEmitterMixin`
https://github.com/ckeditor/ckeditor5-ui/blob/master/src/domemittermixin.js#L70

I need this option in `FocusManager` for listening `focus` and `blur` events which don't bubble. Unfortunately `focusin` and `focusout` (the bubbling ones) are not supported by FF.

oskarwrobel:
Fixes #76

### DIFF
--- a/src/domemittermixin.js
+++ b/src/domemittermixin.js
@@ -10,8 +10,149 @@ import log from '../utils/log.js';
 import isNative from '../utils/lib/lodash/isNative.js';
 
 /**
+ * Mixin that injects the DOM events API into its host. It provides the API
+ * compatible with {@link utils.EmitterMixin}.
+ *
+ * DOM emitter mixin is by default available in the {@link ui.View} class,
+ * but it can also be mixed into any other class:
+ *
+ *		import mix from '../utils/mix.js';
+ *		import DOMEmitterMixin from '../ui/domemittermixin.js';
+ *
+ *		class SomeView {}
+ *		mix( SomeView, DOMEmitterMixin );
+ *
+ *		const view = new SomeView();
+ *		view.listenTo( domElement, ( evt, domEvt ) => {
+ *			console.log( evt, domEvt );
+ *		} );
+ *
+ * @mixin ui.DOMEmitterMixin
+ * @mixes utils.EmitterMixin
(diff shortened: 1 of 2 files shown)


src/domemittermixin.js
```diff
@@ -52,10 +52,12 @@ extend( ProxyEmitter.prototype, EmitterMixin, {
...
 	 * @param {Object} [options.context] The object that represents `this` in the callback. Defaults to the object firing the event.
+	 * @param {Boolean} [options.useCapture=false] Indicates that events of this type will be dispatched to the registered
+	 * listener before being dispatched to any EventTarget beneath it in the DOM tree.
 	 *
 	 * @method ui.ProxyEmitter#on
 	 */
-	on( event ) {
+	on( event, callback, options = {} ) {
 		// Execute parent class method first.
 		EmitterMixin.on.apply( this, arguments );
```
Reinmar: Don't use `arguments` if you don't have to.

Reinmar: By commenting on this now I mean that we can easily fix it. You could also do the same with `listenTo` in fact.

oskarwrobel:
There is a lack of `on()` test at all. Hm... I think we we can test `on()` instead of `listenTo()` and then test if `listenTo()` uses `on()` with proper emitter.

oskarwrobel:
There is also no tests for `off()` method. I'll handle it.

oskarwrobel:
Ok I understand why there is no tests for `on()` and `off()` methods. They can't be directly called in DOM context. Only `listenTo()` and `stopListening()` can call them in DOM context. Direct call of `on()` and `off()` if fully tested in `EmitterMixin` tests.

Reinmar:
I don't understand. Can't we check that a listener added using `on()` is executed when we fire a native event (I saw we can do this)?

oskarwrobel:
To test `element.on()` we need to mix this mixin into some native element but I'm afraid it won't work.

Reinmar:
Yeah, that's a bad idea.

PR #87, opened 2016-10-11
Removed model from View class

### LINKED ISSUE #73: Why do we need view models?
OK, I hate to say that, but I don't remember or I don't know the answer. Either way, it sounds like we overcomplicated this thing...

So, currently we say – the view model (created automatically by each view and used by controllers to push data into the views and get the feedback) is the view's interface. It's how the world communicates with the view.

That's even fine, but:

* It looks like we haven't followed that rule: https://github.com/ckeditor/ckeditor5-ui-default/issues/70 and I also don't know how to make richer methods (like those `attachTo()` – I guess its params should be put in the view model...).
* If the view model is the views public API, then what's left in the view itself? What does its public API represent? We've got there methods like `init()` and `destroy()` plus couple of other things, but still... why moving the rest outside?

oskarwrobel:
Fixes #73 

Related to:
- https://github.com/ckeditor/ckeditor5-ui-default/issues/94
- https://github.com/ckeditor/ckeditor5-link/issues/50

### DIFF
--- a/src/template.js
+++ b/src/template.js
@@ -177,10 +177,10 @@ export default class Template {
 			 *			on: {
 			 *				click: [
 			 *					// "clicked" event will be fired on `observableInstance` when "click" fires in DOM.
-			 *					bind( 'clicked' ),
+			 *					bind.to( 'clicked' ),
 			 *
 			 *					// A custom callback function will be executed when "click" fires in DOM.
-			 *					bind( () => {
+			 *					bind.to( () => {
 			 *						...
 			 *					} )
 			 *				]
@@ -195,7 +195,8 @@ export default class Template {
 			 *
 			 * @static
 			 * @method ui.Template.bind#to
-			 * @param {String} eventNameOrFunctionOrAttribute Name of {@link utils.ObservableMixin} used in the binding.
+			 * @param {String|Function} eventNameOrFunctionOrAttribute An attribute name of {@link utils.ObservableMixin} or a DOM
+			 * event name or an event callback.
 			 * @param {Function} [callback] Allows processing of the value. Accepts `Node` and `value` as arguments.
 			 * @return {ui.TemplateBinding}
 			 */
--- a/src/template.js
+++ b/src/template.js
@@ -195,15 +195,15 @@ export default class Template {
 			 *
 			 * @static
 			 * @method ui.Template.bind#to
-			 * @param {String} attribute Name of {@link utils.ObservableMixin} used in the binding.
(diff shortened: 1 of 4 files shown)


src/template.js
```diff
@@ -195,15 +195,15 @@ export default class Template {
 			 *
 			 * @static
 			 * @method ui.Template.bind#to
-			 * @param {String} attribute Name of {@link utils.ObservableMixin} used in the binding.
+			 * @param {String} eventNameOrFunctionOrAttribute Name of {@link utils.ObservableMixin} used in the binding.
```
oleq: Now it needs explanation, how it works in the case of DOM attributes (attribute) and DOM events (event name or fn).

src/view.js
```diff
@@ -267,6 +260,7 @@ export default class View {
 }
 
 mix( View, DOMEmitterMixin );
+mix( View, ObservableMixin );
```
oleq: Why do we need this? https://github.com/ckeditor/ckeditor5-ui/blob/t/73/src/domemittermixin.js#L33 since
```js
const DOMEmitterMixin = extend( {}, EmitterMixin, ... )
```

src/view.js
```diff
@@ -113,19 +107,18 @@ export default class View {
...
-	 * Cached {@link ui.Template#bind} object is stored in {@link ui.View.#_bind}.
+	 * Cached {@link ui.Template#bind} object is stored in {@link ui.View.#_templateBind}.
 	 *
-	 * @method ui.View#bind
+	 * @method ui.View#templateBind
 	 */
-	get bind() {
-		if ( this._bind ) {
-			return this._bind;
+	get templateBind() {
```
oleq: I'm not sure whether this should be `templateBind` or `bindTemplate`. WDYT?

src/view.js
oskarwrobel: `ObservableMixin` provides `set`, `bind` and `unbind` methods.

Reinmar: OM also provides the EM, so I'd mix it first into the view, before DEM.

oleq:
> Why do we need this? https://github.com/ckeditor/ckeditor5-ui/blob/t/73/src/domemittermixin.js#L33 since

I meant, do we need it at all?

oleq:
After these changes I got

```
classic:145 TypeError: emitter.on is not a function
    at StickyToolbarView.listenTo (emittermixin.js:182)
    at StickyToolbarView.init (stickytoolbarview.js:196)
```

in tests/editor-classic/manual/classic.

Reinmar:
Oops. Sorry :).

PR #89, opened 2016-10-19
t/88: Allow using Views directly in TemplateDefinition

### LINKED ISSUE #88: Allow using Views directly in TemplateDefinition
A follow-up of https://github.com/ckeditor/ckeditor5-ui-default/issues/67.

----

To speed–up the process of creating the UI structure, passing views directly into template definition like this should be allowed:

```js
constructor( locale ) {
    super( locale );

    // TODO we still need to specialise those buttons (their type).
    this.okButton = new ButtonView( locale );
    this.cancelButton = new ButtonView( locale );
    this.unlinkButton = new ButtonView( locale );
    this.urlInput = new LabeledInputView( locale );

        tag: 'div',
        attributes: [
            class: [ 'ck-link-form' ]
        ],
        children: [
            this.urlInput,
            {
                tag: 'div',
                attributes: {
                    class: [ 'ck-link-form-footer' ]
                },
                children: [
                    this.okButton,
                    this.unlinkButton
                ]
            },
            this.unlinkButton
        ]
    } );
}
```

oleq:
Closes #88.

### DIFF
--- a/src/controller.js
+++ b/src/controller.js
@@ -9,7 +9,7 @@ import CKEditorError from '../utils/ckeditorerror.js';
 import EmitterMixin from '../utils/emittermixin.js';
 import mix from '../utils/mix.js';
 
-const anon = '_anonymous';
+const anon = '$anonymous';
 
 /**
  * Basic Controller class.
@@ -56,14 +56,21 @@ export default class Controller {
 			idProperty: 'name'
 		} );
 
+		/**
+		 * Anonymous collection of this controller instance. It groups child controllers
+		 * which are not to be handled by `Controller#collections`–to–`View#region`
+		 * automation. It also means their views must be handled individually
+		 * by the view, i.e. passed as members of {@link ui.TemplateDefinition#children}.
+		 *
+		 * @protected
+		 * @member {ui.ControllerCollection} ui.Controller#_anonymousCollection
+		 */
+		this.collections.add( this._anonymousCollection = new ControllerCollection( anon ) );
+
 		// Listen to {@link ControllerCollection#add} and {@link ControllerCollection#remove}
 		// of newly added Collection to synchronize this controller's view and children
 		// controllers' views in the future.
 		this.collections.on( 'add', ( evt, collection ) => {
-			if ( isAnonymous( collection ) ) {
-				return;
-			}
-
 			// Set the {@link ControllerCollection#parent} to this controller.
 			// It allows the collection to determine the {@link #ready} state of this controller
 			// and accordingly initialize a child controller when added.
@@ -249,6 +256,9 @@ export default class Controller {
 
 		for ( collection of this.collections ) {
 			for ( childController of collection ) {
+				// Anonymous collection {@link ui.Controller#_anonymousCollection} does not allow
+				// automated controller-to-view binding, because there's no such thing as
+				// anonymous Region in the View instance.
 				if ( !isAnonymous( collection ) && this.view && childController.view ) {
 					this.view.regions.get( collection.name ).views.add( childController.view );
 				}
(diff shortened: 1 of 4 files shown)


oleq:
Just FYI: I smuggled a very nice refactoring to the Template class in this PR (https://github.com/ckeditor/ckeditor5-ui/pull/89/commits/80c6f60eef57b3c8fbdd90eb666bb7fd657e624e) ;-)

src/controller.js
```diff
@@ -226,10 +259,39 @@ export default class Controller {
...
+	/**
+	 * Anonymous collection of this controller instance. It groups child controllers
+	 * which are not to be handled by `Controller#collections`–to–`View#region`
+	 * automation. It also means their views must be handled individually
+	 * by the view, i.e. passed as members of {@link ui.TemplateDefinition#children}.
+	 *
+	 * @protected
+	 * @type {ui.ControllerCollection}
+	 */
+	get _anonymousCollection() {
```
Reinmar: Can't we always create it to KISS? Also, we won't need that condition in `collection#add` listener, because then we can add this collection beforehand.

src/template.js
```diff
@@ -513,12 +510,21 @@ export default class Template {
 	 */
 	_renderElementChildren( elOrDocFragment, shouldApply ) {
 		let childIndex = 0;
+		let tpl, rendered;
+
+		for ( let child of this.children ) {
+			tpl = isView( child ) ? child.template : child;
```
Reinmar: I'm not sure what it does... but why is parent view accessing the child view? Cannot it simply use the child's `element` property?

src/template.js
```diff
@@ -571,25 +569,155 @@ export default class Template {
...
+		 */
+
+		/**
+		 * An {@link utils.EmitterMixin} instance used by the binding
+		 * to (either):
+		 *
+		 * * Listen to the attribute change in the {@link ui.TemplateBinding#observable}.
+		 * * Listen to the event in the DOM.
+		 *
+		 * @member {utils.EmitterMixin} ui.TemplateBinding#emitter
```
Reinmar: Should be `utils.Emitter`. Mixin is an object, `Emitter` is an interface.

src/template.js
```diff
@@ -571,25 +569,155 @@ export default class Template {
...
+	 * @param {ui.TemplateBindingDefinition} def The definition of the binding.
+	 */
+	constructor( def ) {
+		Object.assign( this, def );
+
+		/**
+		 * An observable instance of the binding. It provides the attribute
+		 * with the value or passes the event when a corresponding DOM event is fired.
+		 *
+		 * @member {utils.ObservableMixin} ui.TemplateBinding#observable
```
Reinmar: The same as in case of the next property.

src/template.js
```diff
@@ -571,25 +569,155 @@ export default class Template {
...
+		this.emitter.listenTo( this.observable, 'change:' + this.attribute, () => {
+			syncValueSchemaValue( valueSchema, node, updater );
+		} );
+	}
+}
+
+/**
+ * Describes either
+ *
+ * * A binding to {@link utils.ObservableMixin},
```
Reinmar: lower case if it's a part of a sentence :P

src/controller.js
```diff
@@ -216,7 +249,7 @@ export default class Controller {
 
 		for ( collection of this.collections ) {
 			for ( childController of collection ) {
-				if ( this.view && childController.view ) {
+				if ( !isAnonymous( collection ) && this.view && childController.view ) {
```
Reinmar: Would deserve some comment, cause it's not very simple logic.

src/controller.js
```diff
@@ -58,6 +60,10 @@ export default class Controller {
 		// of newly added Collection to synchronize this controller's view and children
 		// controllers' views in the future.
 		this.collections.on( 'add', ( evt, collection ) => {
+			if ( isAnonymous( collection ) ) {
```
Reinmar: As I mentioned later on the code – this can be avoided.

src/controller.js
```diff
@@ -9,6 +9,8 @@ import CKEditorError from '../utils/ckeditorerror.js';
 import EmitterMixin from '../utils/emittermixin.js';
 import mix from '../utils/mix.js';
 
+const anon = '_anonymous';
```
Reinmar: We use `$` to prefix special kind of "items" in many places in the engine. I feel that this anonymous collection is more special than private, so it should follow this convention.

PR #111, opened 2016-11-21
t/92: Fixed: A View#id Number breaks parent ViewCollection.

### LINKED ISSUE #93: Problem with id property in view class
`id` property could be a [common property](https://github.com/ckeditor/ckeditor5-ui-default/blob/master/src/inputtext/inputtextview.js#L34) for input components, but using this property can blow up the editor. It happens when `View` instance is beeing [added to the views collection](https://github.com/ckeditor/ckeditor5-ui/blob/master/src/controller.js#L220). This is because `id` property is [reserved](https://github.com/ckeditor/ckeditor5-utils/blob/master/src/collection.js#L55) for collection item and has to be a `String`.

oleq:
Fixes #93.

### DIFF
--- a/src/viewcollection.js
+++ b/src/viewcollection.js
@@ -21,7 +21,11 @@ export default class ViewCollection extends Collection {
 	 * @param {utils.Locale} [locale] The {@link core.editor.Editor#locale editor's locale} instance.
 	 */
 	constructor( locale ) {
-		super();
+		super( {
+			// An #id Number attribute should be legal and not break the `ViewCollection` instance.
+			// https://github.com/ckeditor/ckeditor5-ui/issues/93
+			idProperty: 'viewUid'
+		} );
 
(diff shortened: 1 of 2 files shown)


src/viewcollection.js
```diff
@@ -70,6 +70,10 @@ export default class ViewCollection extends Collection {
 		 * @member {HTMLElement} ui.ViewCollection#_boundItemsToViewsMap
 		 */
 		this._boundItemsToViewsMap = new Map();
+
+		// An #id Number attribute should be legal and not break the `ViewCollection` instance.
+		// https://github.com/ckeditor/ckeditor5-ui/issues/93
+		this._idProperty = 'viewUid';
```
Reinmar: It's private property. Specify it via the constructor (`options.idProperty`).

PR #120, opened 2017-01-03
t/119: New tooltip architecture

oleq:
Closes https://github.com/ckeditor/ckeditor5-ui/issues/119.
Closes https://github.com/ckeditor/ckeditor5-ui-default/issues/147.
Closes https://github.com/ckeditor/ckeditor5-ui-default/issues/148.
Closes https://github.com/ckeditor/ckeditor5-ui-default/issues/150.

### DIFF
--- a/src/button/buttonview.js
+++ b/src/button/buttonview.js
@@ -45,16 +45,15 @@ export default class ButtonView extends View {
 		/**
 		 * (Optional) Tooltip of the button, i.e. displayed when hovering the button with the mouse cursor.
 		 *
-		 *   * If defined as a `Boolean` (e.g. `true`), then combination of `label` and `keystroke` will be set as tooltip.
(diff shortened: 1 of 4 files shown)


Reinmar:
Is it correct that tooltip isn't displayed for disabled buttons?

oleq:
Yes. It's by design. In fact, I was wondering if this is correct but decided that less distraction is better. The button isn't available anyway. OTOH, I'm aware that this is a blow to the feature discoverability.

Reinmar:
I think it's a bit strange when e.g. you have disabled undo or redo buttons (redo is disabled most of the time, while undo isn't).

oleq:
Let's find out what people will say. It's a single line of CSS to re–enable them.

fredck:
Please re-enable them. This is the default behavior with "title" and users must have a clue on what is that disabled button about.

PR #143, opened 2017-01-17
t/138: Basic a11y support w/ keyboard

### LINKED ISSUE #138: Basic accessibility support for the keyboard
Things to consider:

1. **Critical**: Focusing and navigating the toolbar, as in v4:

   > Press Alt+F10 to navigate to the toolbar. Move to the next and previous toolbar group with TAB and SHIFT+TAB. Move to the next and previous toolbar button with RIGHT ARROW or LEFT ARROW. Press SPACE or ENTER to activate the toolbar button.

   We don't have toolbar groups yet, so there's one thing less to worry about.

1. **Critical**: Focus cycling in forms and balloon panel navigation (i.e. in the link panel), similar to dialog navigation in v4:

    > Inside a dialog, press TAB to navigate to the next dialog element, press SHIFT+TAB to move to the previous dialog element, press ENTER to submit the dialog, press ESC to cancel the dialog. [...]

1. **Critical**: Dropdown navigation (i.e. headings dropdown)

    > Inside a list-box, move to next list item with TAB OR DOWN ARROW. Move to previous list item with SHIFT+TAB or UP ARROW. Press SPACE or ENTER to select the list option. Press ESC to close the list-box.

1. Optional: Page Navigation Using the "Tab" Key, as in http://sdk.ckeditor.com/samples/tabindex.html.

1. Optional: Accessibility instructions dialog similar to v4:

    image

Note: ARIA attributes are not in the scope of this issue.

oleq:
Requires https://github.com/ckeditor/ckeditor5-utils/pull/123.
Closes #138.

### DIFF
--- a/src/dropdown/dropdownview.js
+++ b/src/dropdown/dropdownview.js
@@ -9,8 +9,8 @@
 
 import View from '../view';
 import Template from '../template';
-import FocusTracker from 'ckeditor5-utils/src/focustracker';
-import KeystrokeHandler from 'ckeditor5-utils/src/keystrokehandler';
+import FocusTracker from '@ckeditor/ckeditor5-utils/src/focustracker';
+import KeystrokeHandler from '@ckeditor/ckeditor5-utils/src/keystrokehandler';
 
 /**
  * The dropdown view class.
--- a/src/dropdown/dropdownview.js
+++ b/src/dropdown/dropdownview.js
@@ -24,13 +24,24 @@ export default class DropdownView extends View {
 	constructor( locale, buttonView, panelView ) {
 		super( locale );
 
+		// Extend button's template before it's registered as a child of the dropdown because
+		// by doing so, its #element is rendered and any post–render template extension will
+		// not be reflected in DOM.
+		Template.extend( buttonView.template, {
+			attributes: {
+				class: [
+					'ck-dropdown__button'
+				]
+			}
+		} );
+
 		/**
 		 * Button of this dropdown view.
 		 *
 		 * @readonly
 		 * @member {ui.button.ButtonView} #buttonView
 		 */
-		this.addChildren( this.buttonView = buttonView );
+		this.buttonView = buttonView;
 
 		/**
 		 * Panel of this dropdown view.
@@ -38,7 +49,7 @@ export default class DropdownView extends View {
 		 * @readonly
 		 * @member {module:ui/dropdown/dropdownpanelview~DropdownPanelView} #panelView
 		 */
-		this.addChildren( this.panelView = panelView );
+		this.panelView = panelView;
 
 		/**
 		 * Controls whether the dropdown view is open, which also means its
@@ -78,14 +89,6 @@ export default class DropdownView extends View {
 				buttonView,
 				panelView
 			]
-		} );
-
-		Template.extend( buttonView.template, {
-			attributes: {
-				class: [
-					'ck-dropdown__button'
-				]
-			}
 		} );
 
 		// Toggle the the dropdown when it's button has been clicked.
--- a/src/dropdown/dropdownview.js
+++ b/src/dropdown/dropdownview.js
@@ -9,6 +9,8 @@
 
 import View from '../view';
 import Template from '../template';
+import FocusTracker from 'ckeditor5-utils/src/focustracker';
+import KeystrokeHandler from 'ckeditor5-utils/src/keystrokehandler';
 
 /**
  * The dropdown view class.
@@ -46,6 +48,22 @@ export default class DropdownView extends View {
 		 * @member {Boolean} #isOpen
 		 */
 		this.set( 'isOpen', false );
+
+		/**
+		 * Tracks information about DOM focus in the list.
+		 *
+		 * @readonly
+		 * @member {module:utils/focustracker~FocusTracker}
+		 */
+		this.focusTracker = new FocusTracker();
+
+		/**
+		 * Instance of the {@link module:core/keystrokehandler~KeystrokeHandler}.
+		 *
+		 * @readonly
+		 * @member {module:core/keystrokehandler~KeystrokeHandler}
+		 */
+		this.keystrokes = new KeystrokeHandler();
 
 		this.template = new Template( {
 			tag: 'div',
@@ -103,5 +121,52 @@ export default class DropdownView extends View {
 		 * @observable
 		 * @member {Boolean} #withText
 		 */
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	init() {
+		// Listen for keystrokes coming from within #element.
+		this.keystrokes.listenTo( this.element );
+
+		// Register #element in the focus tracker.
+		this.focusTracker.add( this.element );
+
+		const closeDropdown = ( data, cancel ) => {
+			if ( this.isOpen ) {
+				this.buttonView.focus();
+				this.isOpen = false;
+				cancel();
+			}
+		};
+
+		// Open the dropdown panel using the arrow down key, just like with return or space.
+		this.keystrokes.set( 'arrowdown', ( data, cancel ) => {
+			if ( !this.isOpen ) {
+				this.isOpen = true;
+				cancel();
+			}
+		} );
+
+		// Block the right arrow key (until nested dropdowns are implemented).
+		this.keystrokes.set( 'arrowright', ( data, cancel ) => {
+			if ( this.isOpen ) {
+				cancel();
+			}
+		} );
+
+		// Close the dropdown using the arrow left/escape key.
+		this.keystrokes.set( 'arrowleft', closeDropdown );
+		this.keystrokes.set( 'esc', closeDropdown );
+
+		return super.init();
+	}
+
+	/**
+	 * Focuses the {@link #buttonView}.
+	 */
+	focus() {
+		this.buttonView.focus();
 	}
 }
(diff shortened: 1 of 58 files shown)


src/toolbar/toolbarview.js
```diff
@@ -22,8 +25,39 @@ export default class ToolbarView extends View {
...
+		this.keystrokes = new KeystrokeHandler();
+
+		/**
+		 * Helps cycling over focusable {@link #items} in the toolbar.
+		 *
+		 * @readonly
+		 * @protected
+		 * @member {module:ui/focuscycler~FocusCycler}
+		 */
+		this._focusCycler = new FocusCycler( this.items, this.focusTracker );
```
oleq: I wonder if such API wouldn't be better:

```js
new FocusCycler( {
	keystrokeOrigin: [ this.element ], // or KeystrokeHandler instance?
	focusables: items,
	actions: {
		previous: [ 'arrowleft', 'arrowup' ],
		next: [ 'arrowright', 'arrowdown' ]
	}
} );
```

so the cycler would handle the listeners like

```js

+		const focusNext = ( data, cancel ) => {
 +			const nextFocusable = this._focusCycler.next;
 +
 +			if ( nextFocusable ) {
 +				nextFocusable.focus();
 +			}
 +
 +			cancel();
 +		};

 +		// Navigate toolbar items forwards using the arrow right/down key.
 +		this.keystrokes.set( 'arrowright', focusNext );
 +		this.keystrokes.set( 'arrowdown', focusNext );
```
automatically. Hm?

Reinmar: We can go either way – keep the integration outside of the focus cycler and make it easy to use it. E.g. the `focusNext()` functions could be its methods.

Or we can go a step further and also add the `actions` setting. This isn't crucial, but would simplify the code. If you see a repeating pattern that such an option would be useful, then go ahead.

PS. The object format of `FocusCyler` options is ofc better. The `actions` option should not be required, so you can use the FC in both ways.

Reinmar: BTW, I don't understand why the `.next` may return a falsy value? If it's a cycler and there's at least one item then such a situation should not be possible. Or am I missing something?

src/dropdown/dropdownview.js
```diff
@@ -104,4 +122,51 @@ export default class DropdownView extends View {
 		 * @member {Boolean} #withText
 		 */
 	}
+
+	/**
+	 * @inheritDoc
+	 */
+	init() {
```
Reinmar: I'd like to have others opinions on whether this is a logic which can end up in the view. I guess this is the typical discussion how dry the view should be and we should not try to answer this, but rather focus on defining consistent rules for our code base.

So, the question is – do we define such a logic inside a view or pull it out? Make sure to read this method as well as what got into https://github.com/ckeditor/ckeditor5-ui/pull/143/files#diff-87ab2335c9320fb35ff63d466896a203.

cc @pjasiun @oskarwrobel.

src/focuscycler.js
```diff
@@ -0,0 +1,151 @@
...
+	 * @param {module:utils/focustracker~FocusTracker} focusTracker
+	 */
+	constructor( viewCollection, focusTracker ) {
+		/**
+		 * A view collection the cycler operates on.
+		 *
+		 * @readonly
+		 * @member {module:ui/viewcollection~ViewCollection}
+		 */
+		this.viewCollection = viewCollection;
```
Reinmar: What if view collection contains and item which has focusable parts inside it? Is it checked recursively?

The problem I could see is that if we have button groups (which will be the items of the toolbar), then the toolbar may not know how to create the items collection for the focus cycler. I remember we talked about this, but I don't remember the outcome.

src/focuscycler.js
```diff
@@ -0,0 +1,151 @@
...
+	 */
+	get previous() {
+		return this._getFocusableItem( -1 );
+	}
+
+	/**
+	 * An index of the view in the {@link #viewCollection} which is focused according
+	 * to {@link #focusTracker}. `null` when there's no such view.
+	 *
+	 * @member {Number|null} #current
```
Reinmar: It's readonly.

src/focuscycler.js
Reinmar: In other words – is it the FC's worry to get all the items recursively, or should it be done by the caller?

src/dropdown/dropdownview.js
pjasiun: I'm fine with having smart views which do more than just passing events. We may move it to the util method if one wants to create a similar view with this behavior.

src/toolbar/toolbarview.js
oleq: Well yes and no. The `FocusCycler` cycles but when it visits the view that is already focused, it returns `null` to avoid situations when such an element is re–focused. It would simply make no sense. This way, there's no need to check whether `cycler#next` equals current focused, before calling `focus()`.

src/focuscycler.js
oleq: FC gets a collection as an argument. It's up to the view that spawns FC to provide it with a valid collection of focusable items. Let's say if `ToolbarView`'s items are a multi–level structure like `view->[group1, group2, ... groupN]->[itemN1, itemN2, ..., itemNM]`, it should create the collection which is passed to FC in a flat way `[ item11, item12, ..., item1M, item21, item22, ..., item2M, ..., itemNM ]` (provided each group is `M`-long).

IMO, FC should be as generic as possible. We should not hard code too much logic in it. If we really want to make it smart, let's bring a support for functions to help it navigate items in complex, tree-like structures:

```js
new FocusCycler( {
	// Simple collection of toolbar items.
	focusables: <ViewCollection>,

	// Methods to help FC navigate in a complex, tree–like structure
	focusables: {
		next( current, currentIndex ) {
			if ( current is last in group ) {
				return first in the next group;
			}

			...
		},

		previous( current, currentIndex ) {
			...
		}
	}
} );
```

src/dropdown/dropdownview.js
oskarwrobel: This binding fits perfectly to this view and I think there is no chance that someone will need to change it, so it's cool that this view is smart.

src/dropdown/list/createlistdropdown.js
```diff
@@ -53,6 +53,24 @@ export default function createListDropdown( model, locale ) {
...
+	// If the dropdown panel is already open, the arrow down key should
+	// focus the first element in list.
+	dropdownView.keystrokes.set( 'arrowdown', ( data, cancel ) => {
+		if ( dropdownView.isOpen ) {
+			listView.focus();
+			cancel();
+		}
+	} );
+
+	// If the dropdown panel is already open, the arrow down key should
```
Reinmar: Typo. Should be "arrow up".

src/focuscycler.js
```diff
@@ -0,0 +1,273 @@
...
+ * @module ui/focuscycler
+ */
+
+/**
+ * Helps cycling over focusable views in a {@link module:ui/viewcollection~ViewCollection}
+ * when the focus is tracked by {@link module:utils/focustracker~FocusTracker} instance.
+ */
+export default class FocusCycler {
+	/**
+	 * Creates an instance of the focus cycler utility.
```
Reinmar: Please move these code samples to the class itself. They'll be better visible there.

src/focuscycler.js
```diff
@@ -0,0 +1,273 @@
...
+	 * @param {Object} [options.actions]
+	 */
+	constructor( options ) {
+		Object.assign( this, options );
+
+		/**
+		 * A view collection the cycler operates on.
+		 *
+		 * @readonly
+		 * @member {module:utils/collection~Collection} focusables
```
Reinmar: Missing `#` before property names. Repeats couple of times in this PR.

src/focuscycler.js
```diff
@@ -0,0 +1,273 @@
...
+	get next() {
+		return this._getFocusableItem( 1 );
+	}
+
+	/**
+	 * Returns the previous focusable view in {@link #viewCollection} based on {@link #current}.
+	 * `null` if there's none.
+	 *
+	 * @readonly
+	 * @member {module:ui/view~View|null} #next
```
Reinmar: `#previous`

src/focuscycler.js
```diff
@@ -0,0 +1,273 @@
...
+			}
+
+			return focused;
+		} );
+
+		return index;
+	}
+
+	/**
+	 * Focuses the {@link #first}.
```
Reinmar: Could be `Focuses the {@link #first} item.`

src/list/listview.js
```diff
@@ -30,6 +33,42 @@ export default class ListView extends View {
...
+		 * @readonly
+		 * @protected
+		 * @member {module:ui/focuscycler~FocusCycler}
+		 */
+		this._focusCycler = new FocusCycler( {
+			focusables: this.items,
+			focusTracker: this.focusTracker,
+			keystrokeHandler: this.keystrokes,
+			actions: {
+				// Navigate list items backwards using the arrowup key.
```
Reinmar: I don't think that this requires commenting. This is really clear API.

Reinmar:
Some minor docs issues here and there, but other than that it's superb! :)

src/list/listview.js
Reinmar: Well, if you don't agree, @oleq, then you admit that the API is bad ;).

Issue #139, opened 2017-01-11
Lack of exception when child view is not registered using addChildren()


oskarwrobel:
When you create and use button as child view in template, but you don't register it as a child using `this.addChildren()`, everything works fine except icon. Icon doesn't render and there is no exception, console is clear. I know that every child should be registered, but it is super easy to forget about it and there is no error.

oskarwrobel:
Is it possible to register view as child automatically when template is render?

oleq:
It is possible but I don't think it should happen. It would mean that `Template` must know about the `View`, which is not a very good mix.

oleq:
> but it is super easy to forget about it and there is no error.

I think tests should verify it.

oleq:
In theory, it could be `View's` job. At some point, i.e. on `init()`, it would traverse `#template.children` in search for `View` instances and check whether they are registered or not and throw an error if not. 

Or we could drop `addChildren` completely and perform the registration, following the same algorithm that I suggested above.

PR #146, opened 2017-01-23
t/145: Fixed ViewCollection and DOM desynchronise in some cases.

### LINKED ISSUE #145: ViewCollection and DOM desynchronise in some cases
Steps to reproduce:

1. `VC` is used in a DOM template.
2. Template renders.
3. Contents of `VC` change, i.e. views are removed.
4. `VC.init()` is called.

**Expected**: Children of `VC#_parentElement` reflect directly `View#element` of the views in the colleciton, in the count and order.

**Actual**: There are some extra elements in `#_parentElement`, which belong to views, which i.e. have been removed in (3.) but the removal had not been reflected in DOM.

This happens because of `this.ready` conditions in https://github.com/ckeditor/ckeditor5-ui/blob/master/src/viewcollection.js#L37, which freeze the `VC` in DOM until `init()` is called. 

Because of that, if the VC has been rendered in some template before `init()`, a view can be added to the collection, but it's `#element` will not land there in DOM. Similarly, a view can be removed, but its `#element` still remains in DOM. A serious source of bugs, hard to debug.

oleq:
Fixes #145.

### DIFF
--- a/src/dropdown/dropdownview.js
+++ b/src/dropdown/dropdownview.js
@@ -22,6 +22,17 @@ export default class DropdownView extends View {
 	constructor( locale, buttonView, panelView ) {
 		super( locale );
 
+		// Extend button's template before it's registered as a child of the dropdown because
+		// by doing so, its #element is rendered and any post–render template extension will
+		// not be reflected in DOM.
+		Template.extend( buttonView.template, {
+			attributes: {
+				class: [
+					'ck-dropdown__button'
+				]
+			}
+		} );
+
 		/**
 		 * Button of this dropdown view.
 		 *
@@ -60,14 +71,6 @@ export default class DropdownView extends View {
 				buttonView,
 				panelView
 			]
-		} );
-
-		Template.extend( buttonView.template, {
-			attributes: {
-				class: [
-					'ck-dropdown__button'
-				]
-			}
 		} );
 
 		// Toggle the the dropdown when it's button has been clicked.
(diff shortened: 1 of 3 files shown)


src/dropdown/dropdownview.js
```diff
@@ -22,6 +22,17 @@ export default class DropdownView extends View {
 	constructor( locale, buttonView, panelView ) {
 		super( locale );
 
+		// Extend button's template before it's registered as a child of the dropdown because
+		// by doing so, its #element is rendered and any post–render template extension will
+		// not be reflected in DOM.
+		Template.extend( buttonView.template, {
```
oleq: The only drawback of the new approach, aside from some minor performance impact, is that adding a `View` to any collection means immediate rendering view's `#element`, to synchronize the VC in DOM. 

It's important to remember that fact because i.e. extending view's template after that will have no reflection in DOM since the template has already been rendered.

oskarwrobel: There is a chance that someone will modify template in wrong moment and result will be different than expected and there will be no error or warning. I know that `Template` does not know anything about the `View`, bu this is the second place (first: https://github.com/ckeditor/ckeditor5-ui/issues/139) where `Template` can be used incorrectly without any warning.

oleq: There's an easy fix for this, I think. 

We render a `Template` instance only once despite the fact that it's possible to re–render it and use it as a factory. Because of that, we can simply introduce a flag `Template#rendered`, which would be set after `render()` is executed. Then, [`Template.extend()`](https://github.com/ckeditor/ckeditor5-ui/blob/t/145/src/template.js#L234-L236) would check for that flag, and display a warning informing about attempting to extend an instance, which has already been rendered.

WDYT?

oleq: Created an issue https://github.com/ckeditor/ckeditor5-ui/issues/148.

PR #147, opened 2017-01-23
t/139: Implemented automated registration of child View instances referred in View#template

### LINKED ISSUE #139: Lack of exception when child view is not registered using addChildren()

oleq:
Closes #139.

I though I'll be able to deprecate `View#addChildren()` completely but I cannot. [There are cases](https://github.com/ckeditor/ckeditor5-ui/blob/t/139/src/button/buttonview.js#L194-L201) when child views are added after the template has been rendered and there's absolutely no way to detect whether they were registered or not.

Still, it should help developers avoid missing `addChildren()` in many places and simplify the code even further, which sounds good to me.

### DIFF
--- a/src/dropdown/dropdownview.js
+++ b/src/dropdown/dropdownview.js
@@ -28,7 +28,7 @@ export default class DropdownView extends View {
 		 * @readonly
 		 * @member {ui.button.ButtonView} #buttonView
 		 */
-		this.addChildren( this.buttonView = buttonView );
+		this.buttonView = buttonView;
 
 		/**
 		 * Panel of this dropdown view.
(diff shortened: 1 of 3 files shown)


oskarwrobel:
In the link plugin children are added manually: https://github.com/ckeditor/ckeditor5-link/blob/master/src/ui/linkformview.js#L65 and it cause 
```
CKEditorError: collection-add-item-already-exists
```

PR #149, opened 2017-01-24
t/148: Log when user attempts to extend a Template instance which has alread…

### LINKED ISSUE #148: Implement notification when user attempts to extend a Template instance which has already been rendered
A follow–up of https://github.com/ckeditor/ckeditor5-ui/pull/146#discussion_r97310387.

oleq:
…y been rendered.

Closes #148.

### DIFF
--- a/src/template.js
+++ b/src/template.js
@@ -64,7 +64,7 @@ export default class Template {
 		 * @protected
 		 * @member {Boolean}
 		 */
-		this._rendered = false;
+		this._isRendered = false;
 
 		/**
 		 * Tag of this template, i.e. `div`, indicating that the instance will render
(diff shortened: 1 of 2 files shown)


src/template.js
```diff
@@ -56,6 +57,16 @@ export default class Template {
...
 
 		/**
+		 * Indicates whether this particular Template instance has been
+		 * {@link #render rendered}.
+		 *
+		 * @readonly
+		 * @protected
+		 * @member {Boolean}
+		 */
+		this._rendered = false;
```
Reinmar: Should be `_isRendered`. (https://github.com/ckeditor/ckeditor5-design/wiki/Code-Style-Naming-Guidelines#properties-and-variables)

oleq: Then we should also change `View#ready` to `View#isReady`.

PR #156, opened 2017-02-03
t/152: Implemented FloatingToolbarView and Template.revert()

### LINKED ISSUE #152: API extension and improvements for editor-inline
A follow-up of https://github.com/ckeditor/ckeditor5-editor-inline/issues/1.

Some changes are required for editor-inline to work properly:

1. `View.destroy()` should not call `View#element.remove()`. The inline editor is given an existing element in DOM and it definitely should not nuke it. Besides, it turned out that `View` removing its `#element` from DOM on `destroy()` is doing a job of `ViewCollection`; it's obsolete.
2. `Template.apply()` should preserve original values of attributes. The inline editor should not destroy existing classes and so on, it's only borrowing the element for its lifetime and should be only concerned with own classes.
3. Implement `Template.revert()` to clean–up after 2. and revert the element to the same state before `Editor.create()`.

oleq:
### Proposed merge commit message

Feature: Implemented FloatingToolbarView, Implemented Template.revert(). Closes #152.

**TODO** The above message may be outdated. Verify it!


src/toolbar/floating/floatingtoolbarview.js
```diff
@@ -0,0 +1,183 @@
...
+
+const toPx = toUnit( 'px' );
+
+/**
+ * The floating toolbar view class. It floats around the {@link #targetElement}
+ * to remain visible in the viewport.
+ *
+ * @extends module:ui/toolbar/toolbarview~ToolbarView
+ */
+export default class FloatingToolbarView extends ToolbarView {
```
Reinmar: I've got doubts about this class.

1. It fixes that it must be toolbar what floats and nothing more. What if we'd like to add one more element to it?
2. It fixes that the floating toolbar must be floating in this specific way. "Floating" is just a behaviour and we have at least two floating toolbars already.

This behaviour should rather be implemented similarly to the contextual toolbar – as a container for whatever you want to have inside. Just like the balloon.

src/editableui/editableuiview.js
```diff
@@ -60,6 +60,15 @@ export default class EditableUIView extends View {
 		this.set( 'isFocused', false );
 
 		/**
+		 * Indicates whether the view operates on an external {@link #editableElement} or
+		 * rendered it on its own.
+		 *
+		 * @observable
+		 * @member {Boolean} #isRendered
+		 */
+		this.set( 'isRendered', !editableElement );
```
Reinmar: Kinda misleading name, considering the description. You can do better than that :P

src/panel/floating/floatingpanelview.js
```diff
@@ -0,0 +1,194 @@
...
+			fitInViewport: true
+		} );
+
+		Object.assign( this, { top, left } );
+	}
+}
+
+/**
+ * A default set of positioning functions used by the panel view to float
+ * around {@link targetElement}.
```
Reinmar: Incorrect link. Please run `gulp docs` and fix the errors.

src/panel/floating/floatingpanelview.js
```diff
@@ -0,0 +1,194 @@
...
+ *		[ Panel ]
+ *		+----------------+
+ *		| #targetElement |
+ *		+----------------+
+ *
+ * See {@link #_updatePosition}.
+ *
+ * Positioning functions must be compatible with {@link module:utils/dom/position~Position}.
+ *
+ * @member {Object} module:ui/floatingpanel/floatingpanelview~FloatingPanelView#defaultPositions
```
Reinmar: This is a static property so `.defaultPositions` not `#defaultPositions`.

Also, don't link to protected properties from public ones. (`_updatePosition`).

src/template.js
```diff
@@ -102,6 +102,18 @@ export default class Template {
...
+
+		/**
+		 * Data used by {@link #revert} method to restore a node
+		 * to its original state.
+		 *
+		 * See: {@link #apply}.
+		 *
+		 * @protected
+		 * @readonly
+		 * @member {module:ui/template~RenderConfig#renderData}
```
Reinmar: Incorrect notation.

src/template.js
```diff
@@ -156,7 +177,43 @@ export default class Template {
...
+		} );
+
+		return node;
+	}
+
+	/**
+	 * Reverts a template {@link module:ui/template~Template#apply applied} to an existing DOM Node.
+	 *
+	 * @param {Node} node Root node for the template to revert. Must be the same node
+	 * that {@link module:ui/template~Template#apply} has used.
```
Reinmar: Must be the same node?

src/template.js
```diff
@@ -156,7 +177,43 @@ export default class Template {
...
+	 */
+	revert( node ) {
+		if ( !node ) {
```
Reinmar: No point in checking this.

src/template.js
```diff
@@ -156,7 +177,43 @@ export default class Template {
...
+		if ( !this._revertData ) {
+			/**
+			 * Attempting reverting a template which has not been applied yet.
+			 *
+			 * @error ui-template-revert-not-applied
+			 */
+			throw new CKEditorError( 'ui-template-revert-not-applied: Attempting reverting a template which has not been applied yet.' );
+		}
+
+		this._revertNode( node, this._revertData );
```
Reinmar: You do not revert a node. You remove the template from the node or revert template application from node.

src/template.js
```diff
@@ -265,14 +322,12 @@ export default class Template {
 	 * Renders a DOM Node (either `HTMLElement` or `Text`) out of the template.
 	 *
 	 * @protected
-	 * @param {Node} applyNode If specified, this template will be applied to an existing DOM Node.
-	 * @param {Boolean} intoFragment If set, children are rendered into `DocumentFragment`.
-	 * @returns {HTMLElement|Text} A rendered Node.
+	 * @param {module:ui/template~RenderConfig} config Rendering config.
 	 */
-	_renderNode( applyNode, intoFragment ) {
+	_renderNode( config ) {
```
Reinmar: We rather tend to call such state objects "data".

src/template.js
```diff
@@ -1313,3 +1576,37 @@ function isViewCollection( item ) {
  * @param {Function} [callback] Allows processing of the value. Accepts `Node` and `value` as arguments.
  * @return {module:ui/template~TemplateBinding}
  */
+
+/**
+ * The {@link module:ui/template~Template#_renderNode} config.
+ *
+ * @interface module:ui/template~RenderConfig
```
Reinmar: Isn't this private?

Reinmar:
Not much to say here. I found some naming and docs issues. Other than that, I'm a bit worried about the `Template` size and amount of functionality which it brings. Fortunately, it's a simple (i.e. small) piece of API so there's little impact on the rest of the project if we'll need to work on the internals.

In general, are you happy with the stability of apply + remove? Or is it a piece we'll need to polish in the future?

oleq:
I'm quite happy with how `revert()` works. It passes tests and works for the particular use case. Besides, it's not gonna be a piece of API which is used very often. 

And, yes, `Template` will need to be split sooner or later into text renderer, element renderer, attribute renderer, etc. `Template`, much shorter than it is, would then only glue the pieces and give the nice API to the world.

Reinmar:
I've just noticed that there's no manual test for the floating panel.

oleq:
@Reinmar 👉  https://github.com/ckeditor/ckeditor5-ui/issues/161

PR #160, opened 2017-02-28
t/154: Created ToolbarSeparatorView class and getItemsFromConfig utility

### LINKED ISSUE #154: Create separators for toolbars
Cause this is really confusing now:

image

oleq:

Feature: Created ToolbarSeparatorView class and getItemsFromConfig utility. Closes #154.


src/toolbar/utils.js
```diff
@@ -0,0 +1,34 @@
...
+ * @module ui/toolbar/utils
+ */
+
+import ToolbarSeparatorView from './toolbarseparatorview';
+
+/**
+ * An utility which expands a plain toolbar configuration into a collection
+ * of {@link module:ui/view~View views} using a given factory.
+ *
+ * @param {Object} config The toolbar config.
```
Reinmar: It's an array.

src/toolbar/utils.js
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+/**
+ * @module ui/toolbar/utils
+ */
+
+import ToolbarSeparatorView from './toolbarseparatorview';
+
+/**
+ * An utility which expands a plain toolbar configuration into a collection
```
Reinmar: A utility

src/toolbar/utils.js
```diff
@@ -0,0 +1,34 @@
+/**
```
Reinmar: I'd rather avoid creating utility modules. We never know if there will ever be any more util :D. So, let's start `ui/toolbar/getitemsfromconfig.js`.

src/toolbar/utils.js
```diff
@@ -0,0 +1,34 @@
...
+ * An utility which expands a plain toolbar configuration into a collection
+ * of {@link module:ui/view~View views} using a given factory.
+ *
+ * @param {Object} config The toolbar config.
+ * @param {module:utils/collection~Collection} collection A collection into which the config
+ * is expanded.
+ * @param {module:ui/componentfactory~ComponentFactory} factory A factory producing toolbar items.
+ * @returns {Promise} A promise resolved when all toolbar items are initialized.
+ */
+export function getItemsFromConfig( config, collection, factory ) {
```
Reinmar: Does it have to return a collection? Can't it be an array? An array would be more flexible, but perhaps there are some limitations there?

Reinmar:
Few details to fix.

src/toolbar/utils.js
oleq: It returns a Promise. I don't quite follow what you mean.

Reinmar: Uh, what I meant that it pushes the items to the collection rather than returning them as an array. Is it necessary?

oleq: It's an util to shorten the code. If it returned an array, it would mean there's more to do in the code which uses the util, especially that:
1. There's no way to put an array into existing collection because we don't support `Collection.add( ...iterable )`.
2. There's no way to create a collection out of an array `new Collection( iterable )`.

Reinmar: But if it returns a promise and pushes into an existing collection it's far less general. While you may not see this now, you may have a problem later if you want to do something where `Collection` doesn't fit.

Besides, you decided to go with `Collection` because `Collection()` doesn't accept an iterable... So because of one missing feature, you changed the API of another public method. This is the wrong way around. First focus on the API, then on improving its usage by adding missing features.

oleq: I decided to go with `Collection` because each toolbar inheriting from `ToolbarView` is `Collection`–oriented in items management. It was an obvious choice that the utility for that specific `ToolbarView` class should work with `Collection` by default.

> While you may not see this now, you may have a problem later if you want to do something where Collection doesn't fit.

Then it's probably not `ToolbarView`. And you need a different utility.

oleq: After a second thought, why not a static method of `ToolbarView`?

Reinmar: > After a second thought, why not a static method of ToolbarView?

Perhaps it can go there too. It may be easier to find.

Reinmar: > Then it's probably not ToolbarView. And you need a different utility.

How can you know that today? I have no idea whatsoever how this method will be used a month from now.

You're making here a choice to bind this method with `Collection` for no important reason. Coupling should be done as late as possible. That's the key to a flexible and reusable API. 

Btw, the API docs is wrong, because this method can't be used with a util collection. It must be used with a view collection. Which rings one more bell – this method is doing more than it says if the collection is already ready, because then the views will also be initialized. Here you have another argument for not coupling these things.

Reinmar:
Just a note – I changed the changelog entry to "Feature: Added support for toolbar item separators. Closes #154."

PR #172, opened 2017-03-20
Added `keepAttachedTo ` method to the BalloonPanelView.

### LINKED ISSUE #170: BalloonPanel doesn't follow its target as the webpage scrolls
Eg Link balloon panel.
mar-16-2017 16-39-41

oskarwrobel:

Feature: Added `keepAttachedTo()` method to the `BalloonPanelView`. Closes #170.

## Follow-ups:
- #181 - Panel sticks out of the container element while scroll


src/panel/balloon/balloonpanelview.js
```diff
@@ -159,6 +159,26 @@ export default class BalloonPanelView extends View {
...
+	 * @param {module:utils/dom/position~Options} options Positioning options compatible with
+	 * {@link module:utils/dom/position~getOptimalPosition}. Default `positions` array is
+	 * {@link module:ui/panel/balloon/balloonpanelview~BalloonPanelView.defaultPositions}.
+	 */
+	stickTo( options ) {
+		// First we need to attach the balloon panel to target element.
+		this.attachTo( options );
+
+		// Then we need to listen to scroll of eny element in the document and update position of the balloon panel.
+		this.listenTo( document, 'scroll', () => this.attachTo( options ), { useCapture: true } );
```
oleq: This `useCapture` that you put in here makes the listener very generic. Because of that, events fired in the `document` will trigger the callback in the capture phase. And since the listener is attached to `document`, every single scroll event coming from whatever element with `overflow: scroll` will trigger it. Such element doesn't even need to have anything in common with CKEditor. `document` aggregates all events with `useCapture` [because it's a root](http://stackoverflow.com/a/10654134/1485219).

For one thing, it's OK. Because the balloon will re–position when elements with `overflow: scroll` will scroll. But still, it's too much. We need to filter out the events that have no impact on balloon panel, which are events comming from elements that are neither:
1. the ancestor of the ck-body (or wherever the balloon is located in DOM),
2. nor the ancestor of the target in `attachTo()`.

[`Node.contains()`](https://developer.mozilla.org/en/docs/Web/API/Node/contains) will help a lot.

tests/panel/balloon/balloonpanelview.js
```diff
@@ -409,6 +409,55 @@ describe( 'BalloonPanelView', () => {
 			} );
 		} );
 	} );
+
+	describe( 'stickTo()', () => {
```
oleq: 1. We need a test to verify that use of `useCapture` [makes sense](https://github.com/ckeditor/ckeditor5-ui/pull/172/files#r107615900) and actually works.

2. We need a manual test for `BalloonPanel` because it's getting quite complex. I think we could develop it on top of http://localhost:8125/ckeditor5-ui/tests/manual/tickets/126/1.html.

src/panel/balloon/balloonpanelview.js
oleq: Yes, that's right.

src/panel/balloon/balloonpanelview.js
```diff
@@ -159,6 +159,26 @@ export default class BalloonPanelView extends View {
...
+	/**
+	 * Works exactly the same as {module:ui/panel/balloon/balloonpanelview~BalloonPanelView.attachTo} with one exception.
+	 * Position of attached panel is constantly updated when any of element in document is scrolled. Thanks to this
+	 * panel always sticks to the target element. See #170.
+	 *
+	 * @param {module:utils/dom/position~Options} options Positioning options compatible with
+	 * {@link module:utils/dom/position~getOptimalPosition}. Default `positions` array is
+	 * {@link module:ui/panel/balloon/balloonpanelview~BalloonPanelView.defaultPositions}.
+	 */
+	stickTo( options ) {
```
oleq: I think we should reference the `attachTo` method in this name because it's essentially the same thing. Like `keepAttachedTo` albo `attachPermanentlyTo`.

This could even become a parameter of `attachTo` options. I wonder which naming/grouping strategy is the best /cc @Reinmar.

src/panel/balloon/balloonpanelview.js
```diff
@@ -162,22 +163,36 @@ export default class BalloonPanelView extends View {
...
+		// Then we need to listen on scroll event of eny element in the document.
+		this.listenTo( document, 'scroll', ( evt, domEvt ) => {
+			// And update position if scrolled element contains related to the balloon elements.
+			if ( domEvt.target.contains( target ) || domEvt.target.contains( limiter ) ) {
+				this.attachTo( options );
+			}
+		}, { useCapture: true } );
+
+		// We need to listen on window resize event and update position.
+		this.listenTo( window, 'resize', () => this.attachTo( options ) );
```
oleq: Please use https://github.com/ckeditor/ckeditor5-utils/blob/master/src/dom/global.js#L1-L26 instead. It helps a lot stubbing things when writing tests.

src/panel/balloon/balloonpanelview.js
```diff
@@ -162,22 +163,36 @@ export default class BalloonPanelView extends View {
...
 
 		// After all we need to clean up the listener.
-		this.once( 'change:isVisible', () => this.stopListening( document, 'scroll' ) );
+		this.once( 'change:isVisible', () => {
+			this.stopListening( document, 'scroll' );
+			this.stopListening( window, 'resize' );
```
oleq: Please use https://github.com/ckeditor/ckeditor5-utils/blob/master/src/dom/global.js#L1-L26 instead. It helps a lot stubbing things when writing tests.

src/panel/balloon/balloonpanelview.js
oskarwrobel: Done.

src/panel/balloon/balloonpanelview.js
oskarwrobel: Done.

src/panel/balloon/balloonpanelview.js
oskarwrobel: I limited updating position to scroll events from elements containing `target` and `limiter` elements.

tests/panel/balloon/balloonpanelview.js
oskarwrobel: I prepared dedicated test for this ticket: http://localhost:8125/ckeditor5-ui/tests/manual/tickets/170/1.html this test compares `attachTo` vs `keepAttachedTo`

src/panel/balloon/balloonpanelview.js
oskarwrobel: Renamed to `keepAttachedTo`

src/panel/balloon/balloonpanelview.js
```diff
@@ -151,14 +151,48 @@ export default class BalloonPanelView extends View {
...
+	 * @param {module:utils/dom/position~Options} options Positioning options compatible with
+	 * {@link module:utils/dom/position~getOptimalPosition}. Default `positions` array is
+	 * {@link module:ui/panel/balloon/balloonpanelview~BalloonPanelView.defaultPositions}.
+	 */
+	keepAttachedTo( options ) {
+		// First we need to attach the balloon panel to the target element.
+		this.attachTo( options );
+
+		const target = options.target;
+		const limiter = options.limiter || defaultLimiterElement;
```
oleq: This line lacks test coverage.

tests/manual/tickets/170/1.html
```diff
@@ -0,0 +1,104 @@
...
+				<p></p>
+				<p></p>
+				<p></p>
+				<p></p>
+				<p></p>
+				<p></p>
+				<p></p>
+				<p></p>
+				<p></p>
+				<p>Balloon is attached to the <strong>TARGET</strong> element.</p>
```
oleq: Please use `position: absolute` or simply large `margin` and drop these paragraphs. It just looks weird :D

tests/manual/tickets/170/1.js
```diff
@@ -0,0 +1,86 @@
...
+/* globals window, document, console:false */
+
+import ClassicEditor from '@ckeditor/ckeditor5-editor-classic/src/classic';
+import ArticlePresets from '@ckeditor/ckeditor5-presets/src/article';
+import BalloonPanelView from '@ckeditor/ckeditor5-ui/src/panel/balloon/balloonpanelview';
+import View from '@ckeditor/ckeditor5-ui/src/view';
+import Template from '@ckeditor/ckeditor5-ui/src/template';
+
+// Content of the balloon panel.
+class BalloonContentView extends View {
```
oleq: KISS. Add simple styles 
```css
.ck-balloon-panel {
    padding: 2em;
}
```

and then `balloonPanel.innerHTML = 'Balloon'`. It should be about what is important in that particular test case, not about balloon content, view nesting and the such. It's just a noise.

tests/panel/balloon/balloonpanelview.js
```diff
@@ -409,6 +404,87 @@ describe( 'BalloonPanelView', () => {
...
+			expect( attachToSpy.calledOnce ).to.true;
+			expect( attachToSpy.lastCall.args[ 0 ] ).to.deep.equal( { target, limiter } );
+
+			window.dispatchEvent( new Event( 'resize' ) );
+
+			expect( attachToSpy.calledTwice ).to.true;
+			expect( attachToSpy.lastCall.args[ 0 ] ).to.deep.equal( { target, limiter } );
+		} );
+
+		it( 'should stop attach when balloon is hidden', () => {
```
oleq: I think this test could be merged with `'should stop attach when balloon is hidden'`.

tests/panel/balloon/balloonpanelview.js
```diff
@@ -409,6 +404,87 @@ describe( 'BalloonPanelView', () => {
...
+			expect( attachToSpy.lastCall.args[ 0 ] ).to.deep.equal( { target, limiter } );
+
+			notRelatedElement.dispatchEvent( new Event( 'scroll' ) );
+
+			// Nothing's changed.
+			expect( attachToSpy.calledThrice ).to.true;
+			expect( attachToSpy.lastCall.args[ 0 ] ).to.deep.equal( { target, limiter } );
+		} );
+
+		it( 'should stop attach when balloon is hidden', () => {
```
oleq: s/attach/attaching,
s/balloon/the balloon,

src/panel/balloon/balloonpanelview.js
```diff
@@ -151,14 +151,48 @@ export default class BalloonPanelView extends View {
...
 
 		const { top, left, name: position } = getOptimalPosition( positionOptions );
 
 		Object.assign( this, { top, left, position } );
 	}
+
+	/**
+	 * Works exactly the same as {module:ui/panel/balloon/balloonpanelview~BalloonPanelView.attachTo} with one exception.
+	 * Position of attached panel is constantly updated when any of parents of the target or limiter elements are scrolled
+	 * or when browser window is resized. Thanks to this panel always sticks to the target element. See #170.
```
oleq: Full link to the issue would be safer. We had code moving between repos and such hashes became outdated more than once already.

oleq:
Afterthought: Scroll/resize listeners should be killed on component `destroy()`.

oskarwrobel:
> Also: I wonder if method renaming would make sense: attachTo() => setPosition(), keepAttachedTo() => attachTo()?

I was thinking about one `attachTo` method which should always update position on scroll and resize.

src/panel/balloon/balloonpanelview.js
oskarwrobel: Done.

src/panel/balloon/balloonpanelview.js
oskarwrobel: Done.

tests/manual/tickets/170/1.html
oskarwrobel: Done.

tests/manual/tickets/170/1.js
oskarwrobel: Done.

tests/panel/balloon/balloonpanelview.js
oskarwrobel: Done.

tests/panel/balloon/balloonpanelview.js
oskarwrobel: Done.

src/panel/balloon/balloonpanelview.js
```diff
@@ -151,14 +151,59 @@ export default class BalloonPanelView extends View {
...
+		// First we need to attach the balloon panel to the target element.
+		this.attachTo( options );
+
+		const target = options.target;
+		const limiter = options.limiter || defaultLimiterElement;
+
+		// Then we need to listen on scroll event of eny element in the document.
+		this.listenTo( global.document, 'scroll', ( evt, domEvt ) => {
+			// And update position if scrolled element contains related to the balloon elements.
+			if ( domEvt.target.contains( target ) || domEvt.target.contains( limiter ) ) {
```
oleq: What would happen if `options.target` or `options.limiter` are `Range` or `ClientRect`? I suppose it's not gonna work and it may throw errors.

For the DOM `Range` we can do it by obtaining range ancestor `Node`. For `ClientRect`, there's nothing we can do because it's an abstract geometric entity.

oskarwrobel: Done

oleq:
OK. It's R+ but since the master is locked we postpone the merge.

PR #177, opened 2017-03-28
Introduced ContextualBalloon class for managing contextual balloons.

### LINKED ISSUE #134: There should be an API to manage the editor balloon
We encountered this issue in two places already and I think that this is really universal.

There's no functional code yet, so you'll have to imagine this:

1. Open the image sample.
2. Focus the image. A balloon appears above the image.
3. Click the 'alt text' icon.
4. A new balloon appears... below the image. The toolbar is still visible. We can hide it, but the positions don't match which is a bigger issue

The same happens if you have a contextual toolbar and open a link balloon. Contextual toolbar is linked to the selection focus. The link balloon behaves totally differently. And here we have to hide and show the toolbar manually too.

I'd like to propose that:

1. We expose a set of two functions to show and hide editor contextual toolbar.
2. That every feature will operate on that balloon instance. This means that the content of the balloon will be replaced by the feature which opens it.

Now, the functions to show and hide the balloon will be pretty simple – they will accept `editor.ui` on which they'll store/get the balloon under some symbol. First call to the `show()` function will create the balloon.

There are at least 2 tricky things which I can see:

1. Position configuration. How to solve the issue that new balloon appears in a different position than the previous one. I think that it can be done like this – if the toolbar is already visible when you call `show()` again, then the positions that were previously used are kept. If not, then the passed one will be used (use - you pass positions to `show()`.
2. There may be some sequences in which the features may lost a balloon or one feature may incorrectly override other feature's balloon. By sequences I mean – you select text, the ctx toolbar appears. Then you click the link button and the link feature calls `show()` again. Now, selection moves somewhere or sth odd happens, and the ctx toolbar decides that it needs to be rendered again and, even though you're still editing the link, now you see the toolbar instead. But I guess these things we'll need to see and synchronised on our own.

There's one more thing which will be possible after we do this – nice transitions between these panels. We can, with time, work on some size animations or whatever we want. Not easy, but possible.

The other advantage is that we'll have "the close on focus losts" or "reposition on resize" logic in one place.

Finally, the only question which remains is that all this makes sense. That there always will be just one balloon at a time (unless you specifically create your own one, which ofc will be doable, cause this is just a component).

oskarwrobel:

Feature: Introduced `ContextualBalloon` plugin for managing contextual balloons. Closes #134.

---

### Follow-ups:
- #178 - Improve ContextualBalloon positioning.
- #185 - Improve ContextualBalloon manual test.
- #179 - Custom action when panel of ContextualBalloon becomes visible.
- #186 - Component to manage a stack of child views.

### DIFF
--- a/src/contextualballoon.js
+++ b/src/contextualballoon.js
@@ -8,6 +8,7 @@
  */
 
 import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
+import BalloonPanelView from './panel/balloon/balloonpanelview';
 
 /**
  * Common contextual balloon of the Editor.
@@ -24,6 +25,14 @@ export default class ContextualBalloon {
 	 */
 	constructor() {
 		/**
+		 * Balloon panel view.
+		 *
+		 * @readonly
+		 * @member {module:ui/panel/balloon/balloonpanelview~BalloonPanelView} #view
+		 */
+		this.view = new BalloonPanelView();
+
+		/**
 		 * Stack of panels injected to the balloon. Last one in the stack is displayed
 		 * as content of {@link module:ui/contextualballoon~ContextualBalloon#view}.
 		 *
@@ -31,12 +40,6 @@ export default class ContextualBalloon {
 		 * @member {Map} #_stack
 		 */
 		this._stack = new Map();
-
-		/**
-		 * Balloon panel view.
-		 *
-		 * @member {module:ui:view~View} #view
-		 */
 	}
 
 	/**
--- a/src/contextualballoon.js
+++ b/src/contextualballoon.js
@@ -73,7 +73,7 @@ export default class ContextualBalloon extends Plugin {
 	 * @param {module:ui/view~View} view
 	 * @returns {Boolean}
 	 */
-	isViewInStack( view ) {
+	hasView( view ) {
 		return this._stack.has( view );
 	}
 
@@ -83,7 +83,7 @@ export default class ContextualBalloon extends Plugin {
 	 * @param {module:ui/contextualballoon~ViewConfig} data Configuration of the view.
 	 */
 	add( data ) {
-		if ( this.isViewInStack( data.view ) ) {
+		if ( this.hasView( data.view ) ) {
 			/**
 			 * Trying to add configuration of the same view more than once.
 			 *
@@ -112,7 +112,7 @@ export default class ContextualBalloon extends Plugin {
 	 * @param {module:ui/view~View} view A view to be removed from the balloon.
 	 */
 	remove( view ) {
-		if ( !this.isViewInStack( view ) ) {
+		if ( !this.hasView( view ) ) {
 			/**
 			 * Trying to remove configuration of the view not defined in the stack.
 			 *
--- a/src/contextualballoon.js
+++ b/src/contextualballoon.js
@@ -58,13 +58,15 @@ export default class ContextualBalloon extends Plugin {
 	}
 
 	/**
-	 * Returns configuration of the currently visible view or `null` when there are no
+	 * Returns the currently visible view or `null` when there are no
 	 * views in the stack.
 	 *
-	 * @returns {module:ui/contextualballoon~ViewConfig|null}
+	 * @returns {module:ui/view~View|null}
 	 */
-	get visible() {
-		return this._stack.get( this.view.content.get( 0 ) ) || null;
+	get visibleView() {
+		const item = this._stack.get( this.view.content.get( 0 ) );
+
+		return item ? item.view : null;
 	}
 
 	/**
@@ -93,9 +95,9 @@ export default class ContextualBalloon extends Plugin {
 		}
 
 		// When adding view to the not empty balloon.
-		if ( this.visible ) {
+		if ( this.visibleView ) {
 			// Remove displayed content from the view.
-			this.view.content.remove( this.visible.view );
+			this.view.content.remove( this.visibleView );
 		}
 
 		// Add new view to the stack.
@@ -122,7 +124,7 @@ export default class ContextualBalloon extends Plugin {
 		}
 
 		// When visible view is being removed.
-		if ( this.visible.view === view ) {
+		if ( this.visibleView === view ) {
 			// We need to remove it from the view content.
 			this.view.content.remove( view );
 
--- a/src/contextualballoon.js
+++ b/src/contextualballoon.js
@@ -65,7 +65,7 @@ export default class ContextualBalloon extends Plugin {
 	/**
 	 * Returns `true` when given view is in the stack otherwise returns `false`.
 	 *
-	 * @param {module:ui:view~View} view
+	 * @param {module:ui/view~View} view
 	 * @returns {Boolean}
 	 */
 	isViewInStack( view ) {
@@ -73,7 +73,7 @@ export default class ContextualBalloon extends Plugin {
 	}
 
 	/**
-	 * Adds view to the stack and makes is visible.
+	 * Adds view to the stack and makes it visible.
 	 *
 	 * @param {module:ui/contextualballoon~ViewConfig} data Configuration of the view.
 	 */
--- a/src/contextualballoon.js
+++ b/src/contextualballoon.js
@@ -7,23 +7,27 @@
  * @module ui/contextualballoon
  */
 
-import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
+import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
 import BalloonPanelView from './panel/balloon/balloonpanelview';
+import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
 
 /**
  * Common contextual balloon of the Editor.
  *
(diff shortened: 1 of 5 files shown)


tests/manual/contextualballoon/contextualballoon.js
```diff
@@ -0,0 +1,283 @@
+/**
```
oskarwrobel: Code of this manual test is no pretty because a lot of various things happens there :)
This code should be reduced when ContextualToolbar will become a CKE5 plugin.

oleq:
I see `TypeError: Cannot read property 'add' of undefined` error in http://localhost:8125/ckeditor5-ui/tests/manual/contextualballoon/contextualballoon.html, unable to play with manual test.

oleq:
Not sure if this is a bug, but reporting anyway:
kapture 2017-04-04 at 10 49 14

oskarwrobel:
Hm... this is not a bug of `ContextualBalloon` class but manual test should be improved.

src/contextualballoon.js
```diff
@@ -0,0 +1,170 @@
...
+		 *
+		 * @private
+		 * @member {Map} #_stack
+		 */
+		this._stack = new Map();
+
+		/**
+		 * Balloon panel view.
+		 *
+		 * @member {module:ui:view~View} #view
```
oleq: Anyway, any practical reason why not creating the `BalloonPanelView` instance directly at this point but assigning it like `ctxBalloon.view = new BalloonPanelView()` later on instead?

src/contextualballoon.js
```diff
@@ -0,0 +1,170 @@
...
+ */
+
+/**
+ * @module ui/contextualballoon
+ */
+
+import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
+
+/**
+ * Common contextual balloon of the Editor.
```
oleq: How much work would it take to make it a generic utility? I suppose eventually we may end up with other UI components which should manage their children in a similar way. E.g. a dialog, where you edit some fields, then you go deeper to edit some details or internals, then you go back again, etc., without leaving the dialog. Each editing level would be a different child view of the same dialog. If this is a lot of thinking, designing and coding we may think about it as a follow–up, because it's not necessary to MVP.

tests/manual/contextualballoon/contextualballoon.js
```diff
@@ -0,0 +1,292 @@
...
```
oleq: It would be great to see how to the focus behaves when navigating such structure.

src/contextualballoon.js
```diff
@@ -0,0 +1,170 @@
...
+		return this._stack.has( view );
+	}
+
+	/**
+	 * Adds panel to the stack and makes this panel visible.
+	 *
+	 * @param {module:ui/contextualballoon~Panel} panelData Configuration of the panel.
+	 */
+	add( panelData ) {
+		if ( this._stack.get( panelData.view ) ) {
```
oskarwrobel: Yes. I added `isPanelInstack()` at the end, after I wrote this code.

src/contextualballoon.js
```diff
@@ -0,0 +1,170 @@
...
+	 */
+	constructor() {
+		/**
+		 * Stack of panels injected to the balloon. Last one in the stack is displayed
+		 * as content of {@link module:ui/contextualballoon~ContextualBalloon#view}.
+		 *
+		 * @private
+		 * @member {Map} #_stack
+		 */
+		this._stack = new Map();
```
oskarwrobel: Why not `@private`?

src/contextualballoon.js
oskarwrobel: I moved it to the constructor.

tests/manual/contextualballoon/contextualballoon.js
oskarwrobel: Focus behave is a matter of feature implementation. In this PR https://github.com/ckeditor/ckeditor5-link/pull/92 balloon uses focus cycler and works fine.

src/contextualballoon.js
```diff
@@ -0,0 +1,170 @@
...
+ * @module ui/contextualballoon
+ */
+
+import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
+
+/**
+ * Common contextual balloon of the Editor.
+ *
+ * This class reuses the same {module:ui/view~View} for each contextual balloon panel in the editor UI, makes
+ * possible to add multiple panels to the same balloon (stored in the stack, last one in the stack is visible)
```
oskarwrobel: Renamed.

tests/manual/contextualballoon/contextualballoon.js
oskarwrobel: I created issue about simplifying/improving manual test. https://github.com/ckeditor/ckeditor5-ui/issues/185

tests/manual/contextualballoon/contextualballoon.js
oskarwrobel: See: #185

src/contextualballoon.js
oskarwrobel: I created follow-up #186

src/contextualballoon.js
oleq: A good point :P

src/contextualballoon.js
```diff
@@ -0,0 +1,193 @@
...
+	 * @returns {module:ui/contextualballoon~ViewConfig|null}
+	 */
+	get visible() {
+		return this._stack.get( this.view.content.get( 0 ) ) || null;
+	}
+
+	/**
+	 * Returns `true` when given view is in the stack otherwise returns `false`.
+	 *
+	 * @param {module:ui:view~View} view
```
oleq: Invalid link.

src/contextualballoon.js
```diff
@@ -0,0 +1,193 @@
...
+	 * @returns {Boolean}
+	 */
+	isViewInStack( view ) {
+		return this._stack.has( view );
+	}
+
+	/**
+	 * Adds view to the stack and makes is visible.
```
oleq: s/is/it

src/contextualballoon.js
```diff
@@ -0,0 +1,197 @@
...
+import BalloonPanelView from './panel/balloon/balloonpanelview';
+import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
+
+/**
+ * Provides the common contextual balloon panel for the editor.
+ *
+ * This plugin allows reusing a single {module:ui/panel/balloon/balloonpanelview~BalloonPanelView} instance
+ * to display multiple contextual balloon panels in the editor.
+ *
+ * Child views of such a panel are stored in the stack and the last one in the stack is visible. When the
```
oleq: An afterthought: Since `_stack` is private we should not talk in the public docs about it. It is of no concern to the devs how the CB handles views.

src/contextualballoon.js
```diff
@@ -0,0 +1,197 @@
...
+		return this._stack.get( this.view.content.get( 0 ) ) || null;
+	}
+
+	/**
+	 * Returns `true` when the given view is in the stack. Otherwise returns `false`.
+	 *
+	 * @param {module:ui/view~View} view
+	 * @returns {Boolean}
+	 */
+	isViewInStack( view ) {
```
oleq: I would rename it to `hasView()`. As in https://github.com/ckeditor/ckeditor5-ui/pull/177/files#r109897538, `_stack` is private and it does not concern people in the public API.

src/contextualballoon.js
oskarwrobel: When I wrote about stack I wasn't thinking about var name but about the way how we store it.

src/contextualballoon.js
```diff
@@ -0,0 +1,197 @@
...
+		return 'contextualballoon';
+	}
+
+	/**
+	 * Returns configuration of the currently visible view or `null` when there are no
+	 * views in the stack.
+	 *
+	 * @returns {module:ui/contextualballoon~ViewConfig|null}
+	 */
+	get visible() {
```
oleq: > Boolean properties and variables are always prefixed by an auxiliary verb:

https://github.com/ckeditor/ckeditor5-design/wiki/Code-Style-Naming-Guidelines

oskarwrobel: This getter returns visible view. It's not checking if given view is visible.

oleq: So `get visibleView()` maybe? I'm trying to make it less ambiguous.

oskarwrobel: Hm... this returns on only the view but the whole configuration `{ view, position }` https://github.com/ckeditor/ckeditor5-ui/pull/177/files#diff-93d63500d90d723217fc16438b6b32a4R193

oskarwrobel: But maybe only the view should be returned.

oskarwrobel: Hm... looks like there is no need to return the balloon position configuration because this position is used internally by ContextualBalloon.

oskarwrobel: Done.

oleq:
R+ but the master is locked so we're gonna postpone the merge.

PR #188, opened 2017-04-07
T/182 Introduced contextual toolbar plugin

### LINKED ISSUE #182: Introduce contextual toolbar component
See: https://github.com/ckeditor/ckeditor5/issues/402

### LINKED ISSUE #187: Incorrect position of toolbar on the first show
Follow-up: #182 

apr-07-2017 10-46-25

oskarwrobel:

Feature: Introduced `ContextualToolbar` plugin. Closes #182. Closes #187.

Introduced four new `BalloonPanelView` default positions. Added `className` property to the `BalloonPanelView` interface.

BREAKING CHANGE: Default positions of `BalloonPanelView` have been renamed.

---

### Follow-ups:
- ~https://github.com/ckeditor/ckeditor5-ui/issues/187~ - ~Incorrect position of toolbar on the first show~.
- #195 - Balloon opened from contextual toolbar doesn't update its position on resize/scroll.
- #207 - Keyboard support for the ContextualToolbar.


src/toolbar/contextualtoolbar.js
```diff
@@ -0,0 +1,251 @@
+/**
```
oskarwrobel: I'm not sure if it's a good place for this plugin.

oskarwrobel:
Needs to be updated.

src/contextualballoon.js
```diff
@@ -167,7 +177,13 @@ export default class ContextualBalloon extends Plugin {
 	 */
 	_show( view ) {
 		this.view.content.add( view );
-		this.view.attachTo( this._getBalloonPosition() );
+
+		// When view is not rendered we need to wait for it. See: https://github.com/ckeditor/ckeditor5-ui/issues/187.
```
oskarwrobel: This should be 
```
return this.view.content.add( view ).then( () => {
    this.view.attachTo( this._getBalloonPosition() );
} );
```
but I'm gotin to impreve this in: https://github.com/ckeditor/ckeditor5-ui/issues/200

src/toolbar/contextualtoolbar.js
```diff
@@ -0,0 +1,258 @@
...
+const positions = {
+	//     [text range]
+	//                ^
+	//       +-----------------+
+	//       |     Balloon     |
+	//       +-----------------+
+	forwardSelection: ( targetRect, balloonRect ) => ( {
+		top: targetRect.bottom + arrowVOffset,
+		left: targetRect.right - balloonRect.width / 2,
+		name: 's'
```
oskarwrobel: Position names should be updated.

oleq: I think these positions should become part of [BPV's defaults](https://github.com/ckeditor/ckeditor5-ui/blob/t/182/src/panel/balloon/balloonpanelview.js#L371) because there's pretty common (aren't they?) but not necessarily [used by default](https://github.com/ckeditor/ckeditor5-ui/blob/t/182/src/panel/balloon/balloonpanelview.js#L173-L178).

tests/toolbar/contextualtoolbar.js
```diff
@@ -0,0 +1,377 @@
...
+		// Collapse range silently (without firing `change:range` { directChange: true } event).
+		const range = editor.document.selection._ranges[ 0 ];
+		range.end = range.start;
+
+		editor.document.selection.fire( 'change:range', { directChange: false } );
+
+		expect( balloon.visibleView ).to.null;
+	} );
+
+	it( 'should open below if the selection is forward', () => {
```
oleq: This test fails in my local env. Not sure why.

tests/toolbar/contextualtoolbar.js
```diff
@@ -0,0 +1,377 @@
...
+			left: 0,
+			width: 1000,
+			top: 0,
+			height: 1000
+		} );
+
+		contextualToolbar.fire( '_selectionChangeDone' );
+
+		expect( balloon.visibleView ).to.equal( contextualToolbar.toolbarView );
+		expect( balloon.view.top ).to.be.above( 310 );
```
oleq: Please avoid this kind of assertions. Simply check whether the position name is correct, like [this](https://github.com/ckeditor/ckeditor5-ui/blob/t/182/tests/panel/balloon/balloonpanelview.js#L355) and then, separately, check whether positioning functions are correct. This way we'll avoid fixing maaaany magic numbers like 310 in dozens of tests if the positioning function gets improved/changed in the future even by `1px`.

We should simply make sure each of `BalloonPanelView.defaultPositions` does its math and then assert just position names.

oleq: I see `BalloonPanelView.defaultPositions` has no unit tests but this occasion is as good as any other to do it and check `BalloonPanelView` for magic number assertions too.

src/toolbar/contextualtoolbar.js
```diff
@@ -0,0 +1,258 @@
...
+		 * This is to makes easy test debounced action without need to use `setTimeout`. Lodash keeps time related
+		 * stuff in a closure and it's not possible to override it by sinon fake timers.
+		 *
+		 * This debounced function is stored as a plugin property to make possible to cancel
+		 * trailing debounced invocation on destroy.
+		 *
+		 * @private
+		 * @member {Function}
+		 */
+		this._fireChangeDoneDebounced = debounce( () => this.fire( '_selectionChangeDone' ), 200 );
```
oleq: Missing docs for the event. We could also simply name it `_selectionChange` or `_selectionChangeDebounced`, I think.

src/toolbar/contextualtoolbar.js
```diff
@@ -0,0 +1,258 @@
...
+	/**
+	 * @inheritDoc
+	 */
+	init() {
+		/**
+		 * The toolbar view displayed in the balloon.
+		 *
+		 * @member {module:ui/toolbar/toolbarview~ToolbarView}
+		 */
+		this.toolbarView = new ToolbarView( this.editor.locale );
```
oleq: I see a double border around the toolbar + the toolbar has an incorrect background color in http://localhost:8125/ckeditor5-ui/tests/manual/contextualtoolbar/contextualtoolbar.html. There should be a theme file to manage it.

oleq: I think `.ck-toolbar__container` is the right class to use.

oskarwrobel:
Everything is fixed, but Link plugin needs to be aligned to changes in ContextualBalloon so please do not merge.

src/panel/balloon/balloonpanelview.js
```diff
@@ -89,6 +89,15 @@ export default class BalloonPanelView extends View {
 		this.set( 'withArrow', true );
 
 		/**
+		 * Additional css class added to the {#element}.
+		 *
+		 * @observable
+		 * @default ''
+		 * @member {String} #className
+		 */
+		this.set( 'className', '' );
```
oleq: I think `this.set( 'className' );` is enough.

src/panel/balloon/balloonpanelview.js
```diff
@@ -391,5 +437,29 @@ BalloonPanelView.defaultPositions = {
 		top: targetRect.top - balloonRect.height - BalloonPanelView.arrowVerticalOffset,
 		left: targetRect.left + targetRect.width / 2 - balloonRect.width + BalloonPanelView.arrowHorizontalOffset,
 		name: 'arrow_nw'
+	} ),
+
+	forwardSelection: ( targetRect, balloonRect ) => ( {
```
oleq: I'm not sure whether *Selection is the right naming pattern. `BalloonPanelView` has nothing to do with the selection and such position can be used without any range context. It will look weird if someone used it to attach a tooltip or something but the position was named `forwardSelection`. 

We must figure out how to name the positions now that there are more of them. I suggest `[balloon relative placement][arrow placement]` pattern. So `se` becomes `southArrowNorthWest` and `forwardSelection` becomes `southEastArrowNorth`. WDYT? (note that `name` properties won't change as they're in fact CSS classes.)

src/contextualballoon.js
```diff
@@ -178,7 +194,7 @@ export default class ContextualBalloon extends Plugin {
 	 * @returns {module:utils/dom/position~Options}
 	 */
 	_getBalloonPosition() {
-		return Array.from( this._stack.values() )[ 0 ].position;
+		return nth( 0, this._stack )[ 1 ].position;
```
oleq: I see that `Collection` does not allow non-string keys. Pity. But let's get rid of this `nth()` anyway by using `this._stack.values().next().value.position`.

src/contextualballoon.js
```diff
@@ -151,10 +156,16 @@ export default class ContextualBalloon extends Plugin {
...
-	 * of the first view in the stack.
+	 * Updates the position of the balloon panel according to the given position data
+	 * or position data of the first view in the stack.
+	 *
+	 * @param {module:utils/dom/position~Options} [position] position options.
 	 */
-	updatePosition() {
+	updatePosition( position ) {
+		if ( position ) {
+			nth( 0, this._stack )[ 1 ].position = position;
```
oleq: `this._stack.values().next().value.position`

src/toolbar/contextual/contextualtoolbar.js
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+	/**
+	 * Creates toolbar components based on given configuration.
+	 * This needs to be done when all plugins will be ready.
+	 *
+	 * @inheritDoc
+	 */
+	afterInit() {
+		const config = this.editor.config.get( 'contextualToolbar' );
```
oleq: Why does contextual toolbar have a separate config? Shouldn't it use the default toolbar configuration if it is to become a standard editor implementation like editor-inline or editor-classic?

src/toolbar/contextual/contextualtoolbar.js
```diff
@@ -0,0 +1,220 @@
...
+	 * @inheritDoc
+	 */
+	destroy() {
+		this._fireSelectionChangeDebounced.cancel();
+		this.stopListening();
+		super.destroy();
+	}
+
+	/**
+	 * This is internal plugin event which is fired 200 ms after model selection last change (lodash#debounce).
```
oleq: I think there's no need to mention which utility is behind this debouncing. Whatever we use, it's of no concern to the user.

tests/contextualballoon.js
```diff
@@ -96,12 +97,38 @@ describe( 'ContextualBalloon', () => {
...
 
+		it( 'should wait for view init', () => {
+			balloon.add( {
+				view: viewA,
+				position: { target: 'fake' }
+			} );
+
+			sinon.assert.notCalled( balloon.view.attachTo );
+
+			viewA.ready = true;
```
oleq: Are you sure this is the right way to test it? Wouldn't

```js
return balloon.add().then( () =>  {
   // assert when ready
} );

// assert when not ready
```
make more sense?

src/toolbar/contextual/contextualtoolbar.js
oskarwrobel: Hm... But how in this case create an editor with both main and contextual toolbar with different components?

oskarwrobel:
Unfortunately, tests print a lot of errors on the browser console. The same issue as https://github.com/ckeditor/ckeditor5-link/issues/94. This have no impact on the assertions. To fix it will have to refactor CT similar to https://github.com/ckeditor/ckeditor5-link/pull/105

src/panel/balloon/balloonpanelview.js
```diff
@@ -369,27 +415,51 @@ BalloonPanelView.arrowVerticalOffset = 15;
...
+	northWestArrowSouthEast: ( targetRect, balloonRect ) => ( {
 		top: targetRect.top - balloonRect.height - BalloonPanelView.arrowVerticalOffset,
 		left: targetRect.left + targetRect.width / 2 - balloonRect.width + BalloonPanelView.arrowHorizontalOffset,
 		name: 'arrow_nw'
+	} ),
+
+	southEastArrowNorth: ( targetRect, balloonRect ) => ( {
+		top: targetRect.bottom + BalloonPanelView.arrowVerticalOffset,
+		left: targetRect.right - balloonRect.width / 2,
+		name: 'arrow_s'
```
oskarwrobel: Hm... `arrow_s` is for `^` and `arrow_n` is for `V`. Isn't it opposite?

oleq: Yeah, it's confusing. Gonna change it.

src/toolbar/contextual/contextualtoolbar.js
oleq: A good point.

PR #204, opened 2017-04-13
t/203: ViewCollection#destroy should wait for all ViewCollection#add promises to resolve to avoid errors. Closes #203.

### LINKED ISSUE #203: Destruction of the editor when ViewCollection is adding children may end up in errors
`ViewCollection.add( child )` returns a promise but it doesn't belong to `editor.create()` and `editor.destroy()` chain. So it is possible to destroy the editor when `ViewCollection.add( child )` promise is working (i.e. `child.init()` takes a long time to finish). 

Each promise created by `ViewCollection.add( child )` should be remembered in the property of the `VC` like `this._addPromises = [ promises for each add() ]` and removed from the array once resolved so when `ViewCollection.destroy()` is called, it should wait for `Promise.all( this._addPromises )` to finish.

A follow-up of https://github.com/ckeditor/ckeditor5-link/pull/104.

oleq:

Fix: ViewCollection#destroy should wait for all ViewCollection#add promises to resolve to avoid errors. Closes #203.

### DIFF
--- a/src/view.js
+++ b/src/view.js
@@ -269,15 +269,20 @@ export default class View {
 	 *		}
 	 *
 	 * @param {module:ui/view~View|Iterable.<module:ui/view~View>} children Children views to be registered.
+	 * @returns {Promise}
 	 */
 	addChildren( children ) {
 		if ( !isIterable( children ) ) {
 			children = [ children ];
 		}
 
+		const promises = [];
+
 		for ( let child of children ) {
-			this._unboundChildren.add( child );
+			promises.push( this._unboundChildren.add( child ) );
 		}
+
+		return Promise.all( promises );
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


src/viewcollection.js
```diff
@@ -99,13 +109,19 @@ export default class ViewCollection extends Collection {
...
+		// https://github.com/ckeditor/ckeditor5-ui/issues/203
+		return Promise.all( this._addPromises )
+			// Then begin the process of destroying the children.
+			.then( () => {
+				let destroyPromises = [];
 
-		for ( let view of this ) {
-			promises.push( view.destroy() );
-		}
+				for ( let view of this ) {
```
oskarwrobel: You should use const instead of let because `view` is never reassigned or you can use `Array.from` with build in `map` function instead of iterator.
```
.then( () => {
	return Promise.all( Array.from( this, view => view.destroy() ) );
} );
```

oskarwrobel:
Besides one comment it looks ok.

oskarwrobel:
One test fails.

src/view.js
```diff
@@ -269,15 +269,20 @@ export default class View {
...
 		}
 
+		const promises = [];
+
 		for ( let child of children ) {
-			this._unboundChildren.add( child );
+			promises.push( this._unboundChildren.add( child ) );
 		}
+
+		return Promise.all( promises );
```
oskarwrobel: I prefer to use `Arry.from` instead of creating separate array.
```javascript
return Promise.all( Array.from( children, child => {
    return this._unboundChildren.add( child );
} ) );
```
or shorter (but looks tricky)
```javascript
return Promise.all( Array.from( children, child => this._unboundChildren.add( child ) ) );
```

oskarwrobel:
I left one comment but it's not an issue.

PR #208, opened 2017-04-20
t/206: BalloonPanelView should not be focusable

### LINKED ISSUE #206: BalloonPanelView should not be focusable
kapture 2017-04-19 at 9 09 23

Because of that we end up with problems like https://github.com/ckeditor/ckeditor5-link/issues/90. It's probably just a matter of using the `@mixin ck-unselectable` in SASS.

oleq:

Fix: BalloonPanelView should not be focusable. Closes #206.

---

### Additional information

`BPV` has `@mixin ck-unselectable` enabled by default but CSS `user-select` does not affect focus.

Removing `tabindex="-1"` attribute disabled panel focus but it didn't prevent focus loss, so when clicked the panel didn't get focus but editor focus was lost, which was evident in editor-inline.

Only `preventDefault` of the `mousedown` event disabled focus properly and prevented focus loss.

### DIFF
--- a/src/panel/balloon/balloonpanelview.js
+++ b/src/panel/balloon/balloonpanelview.js
@@ -134,10 +134,12 @@ export default class BalloonPanelView extends View {
 					top: bind.to( 'top', toPx ),
 					left: bind.to( 'left', toPx ),
 					maxWidth: bind.to( 'maxWidth', toPx )
-				},
+				}
+			},
 
-				// Make this element `focusable` to be available for adding to FocusTracker.
-				tabindex: -1
+			on: {
+				// https://github.com/ckeditor/ckeditor5-ui/issues/206
+				mousedown: bind.to( evt => evt.preventDefault() )
 			},
 
 			children: this.content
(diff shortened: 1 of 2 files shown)


oleq:
Additional note for the reviewer: please make sure other features than [Link](https://github.com/ckeditor/ckeditor5-link/issues/90) using `BPV` work properly like editor-inline toolbar, image toolbar, image alt text balloon etc.

oskarwrobel:
I'm not able to select input by a click.

apr-21-2017 09-21-08

src/panel/balloon/balloonpanelview.js
```diff
@@ -134,10 +134,12 @@ export default class BalloonPanelView extends View {
...
 					maxWidth: bind.to( 'maxWidth', toPx )
-				},
+				}
+			},
 
-				// Make this element `focusable` to be available for adding to FocusTracker.
-				tabindex: -1
+			on: {
+				// https://github.com/ckeditor/ckeditor5-ui/issues/206
+				mousedown: bind.to( evt => evt.preventDefault() )
```
oskarwrobel: This makes impossible to select any text on balloon.

oleq: You couldn't do it anyway because of `@mixin ck-unselectable` I think.

oleq: Hm. We should probably filter target then. I'll check this out.

oleq:
A better solution in https://github.com/ckeditor/ckeditor5-ui/pull/213.

PR #217, opened 2017-04-24
Added manual test to check balloons position after external changes.

### LINKED ISSUE #198: Position of balloons should be updated on external changes
The first user clicks on link element and sees balloon panel attached to this element. The second user types something before this link element. The result: link element is moving but attached balloon stays at the same position.

oskarwrobel:

Other: Added manual test testing `ContextualToolbar` position after external changes. Closes #198.

---

FollowUps:
- https://github.com/ckeditor/ckeditor5-link/issues/113 - Link balloon does not update its position on external changes.
- https://github.com/ckeditor/ckeditor5-image/issues/106 - Alt balloon does not update its position on external changes.


Reinmar:
Nice try https://www.youtube.com/watch?v=gpsUCEl9zpE ;)

oleq:
Most certainly the MT needs an update after https://github.com/ckeditor/ckeditor5-image/issues/60
```
 +	image: {
 +		toolbar: [ 'imageStyleFull', 'imageStyleSide', '|' , 'imageTextAlternative' ]
 +	}
```

oskarwrobel:
> Most certainly the MT needs an update after ckeditor/ckeditor5-image#60

Done.

oleq:
I think this PR should not close #198 because:
1. the problem is not fixed in this issue,
2. the problem does not concern ckeditor5-ui so much.

I suppose each test case from this MT should land in the proper repository (link, image). In fact, only `ContextualToolbar` test cases belong in this repository. 

Let's try not to abuse ckeditor5-ui with the feature stuff which doesn't belong here. It's a framework, after all.

PR #226, opened 2017-05-09
Added beforeShow event to the ContextualToolbar plugin

### LINKED ISSUE #222: Prevent ContextualToolbar of being displayed in specific places
I need to prevent `ContextualToolbar` of being displayed in some specific situation (when selection range intersects with some specific node). ATM `CT` displays always when the selection is non-collapsed and there is no way to configure it.

WDYT about adding `beforeShow` event to the `CT` plugin. This event will be fired after selection stops changing just before showing the toolbar. Other plugins will be able to listen to this event and modifies `CT` behavior eg by stopping this event what will prevent `CT` of being displayed?

oskarwrobel:

Feature: Added `beforeShow` event to the `ContextualToolbar` plugin. Closes #222.

### DIFF
--- a/src/toolbar/contextual/contextualtoolbar.js
+++ b/src/toolbar/contextual/contextualtoolbar.js
@@ -75,7 +75,7 @@ export default class ContextualToolbar extends Plugin {
 		 * @private
 		 * @member {Function}
 		 */
-		this._resolveShowPanelPromise = spy();
+		this._showPanelPromiseResolver = spy();
 
 		// Attach lifecycle actions.
 		this._handleSelectionChange();
@@ -145,7 +145,7 @@ export default class ContextualToolbar extends Plugin {
 	 */
 	stop( evt ) {
 		evt.stop();
-		this._resolveShowPanelPromise();
+		this._showPanelPromiseResolver();
 		this.stopListening( this, 'beforeShow' );
 	}
 
@@ -172,7 +172,7 @@ export default class ContextualToolbar extends Plugin {
 		}
 
 		const showPromise = new Promise( ( resolve ) => {
-			this._resolveShowPanelPromise = resolve;
+			this._showPanelPromiseResolver = resolve;
 
 			// If `beforeShow` event is not stopped by any external code then panel will be displayed.
 			this.listenTo( this, 'beforeShow', ( evt ) => {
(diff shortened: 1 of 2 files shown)


src/toolbar/contextual/contextualtoolbar.js
```diff
@@ -147,17 +171,35 @@ export default class ContextualToolbar extends Plugin {
...
+					} )
+				);
+
+				// Stop listening on `beforeShow` event.
+				evt.off();
+			} );
+		}, { priority: 'lowest' } );
+
+		// Fire this event to inform interested plugins that `ContextualToolbar` is going to be shown.
+		this.fire( 'beforeShow' );
```
oleq: Can't we pass the stopping function along this event to avoid the additional `stop` method?

src/toolbar/contextual/contextualtoolbar.js
```diff
@@ -147,17 +171,35 @@ export default class ContextualToolbar extends Plugin {
...
+			this._showPanelPromiseResolver = resolve;
 
-		// Add panel to the common editor contextual balloon.
-		return this._balloon.add( {
-			view: this.toolbarView,
-			position: this._getBalloonPositionData(),
-			balloonClassName: 'ck-toolbar-container'
-		} );
+			// If `beforeShow` event is not stopped by any external code then panel will be displayed.
+			this.listenTo( this, 'beforeShow', ( evt ) => {
```
oleq: Wouldn't `once()` mean that `evt.off()` at the end is gone?

oskarwrobel: `once()` uses `on()` but `on()` is not cleared on `stopListening()` (https://github.com/ckeditor/ckeditor5-utils/issues/144). That's why I used `evt.off()`.

src/toolbar/contextual/contextualtoolbar.js
```diff
@@ -147,17 +150,40 @@ export default class ContextualToolbar extends Plugin {
...
 		}
 
-		// Update panel position when selection changes while balloon will be opened (by a collaboration).
-		this.listenTo( this.editor.editing.view, 'render', () => {
-			this._balloon.updatePosition( this._getBalloonPositionData() );
+		const showPromise = new Promise( ( resolve ) => {
+			// If `beforeShow` event is not stopped by any external code then panel will be displayed.
+			this.once( 'beforeShow', () => {
+				if ( isStopped ) {
+					isStopped = false;
```
oleq: Why do we need to reset the value of it?

oskarwrobel: We don't. Fixing...

PR #233, opened 2017-05-16
t/227: Contextual toolbar should re–position correctly on window scroll

### LINKED ISSUE #227: Contextual toolbar is incorrectly positioned on scroll
may-10-2017 15-14-54

And it blinks ugly.

oleq:

Fix: Contextual toolbar should re–position correctly on window scroll. Closes #227.

---

### Additional information

Requires https://github.com/ckeditor/ckeditor5-utils/issues/157.
Also closes https://github.com/ckeditor/ckeditor5-ui/issues/195.

### DIFF
--- a/tests/toolbar/contextual/contextualtoolbar.js
+++ b/tests/toolbar/contextual/contextualtoolbar.js
@@ -140,7 +140,7 @@ describe( 'ContextualToolbar', () => {
 			editor.editing.view.isFocused = true;
 		} );
 
-		it( 'should return promise', () => {
+		it( 'should return a promise', () => {
 			setData( editor.document, '<paragraph>b[a]r</paragraph>' );
 
 			const returned = contextualToolbar._showPanel();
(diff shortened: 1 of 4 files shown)


tests/toolbar/contextual/contextualtoolbar.js
```diff
@@ -160,7 +160,7 @@ describe( 'ContextualToolbar', () => {
 					view: contextualToolbar.toolbarView,
 					balloonClassName: 'ck-toolbar-container ck-editor-toolbar-container',
 					position: {
-						target: forwardSelectionRect,
```
oleq: Note that tests previously passed probably because the `forwardSelectionRect` and `backwardSelectionRect` look exactly the same. Please double–check it though.

oskarwrobel:
`target` function for `CT` is not called on scroll.

may-16-2017 14-17-49

oleq:
I pushed a commit that should fix it @oskarwrobel. Please let me know if it works for you.

PR #235, opened 2017-05-19
t/228: Clicking and dragging in a dropdown panel should not break the selection

### LINKED ISSUE #228: Misclicks on drop down panels cause selection to be lost
We had a ticket for it, but I can't find it at the moment. I can really often reproduce this issue:

may-10-2017 15-22-57

I think it's related to the issue we had in the past that when there's any dragstart+move sequence the selection is being lost. This is very important issue when you try to actually use the editor for real.

oleq:

Fix: Clicking and dragging in a dropdown panel should not break the selection. Closes #228.

### DIFF
--- a/src/dropdown/dropdownpanelview.js
+++ b/src/dropdown/dropdownpanelview.js
@@ -51,7 +51,12 @@ export default class DropdownPanelView extends View {
 				]
 			},
 
-			children: this.children
+			children: this.children,
+
+			on: {
+				// https://github.com/ckeditor/ckeditor5-ui/issues/228
+				mousedown: bind.to( evt => evt.preventDefault() )
+			}
 		} );
 	}
 }
--- a/src/dropdown/dropdownpanelview.js
+++ b/src/dropdown/dropdownpanelview.js
@@ -55,7 +55,7 @@ export default class DropdownPanelView extends View {
 
 			on: {
 				// https://github.com/ckeditor/ckeditor5-ui/issues/228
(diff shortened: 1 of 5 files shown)


oleq:
Although it fixes the issue I'm not quite sure if it isn't a little bit too much (too aggressive). 

What possible problems could we face (in the future) because of this fix? If we decided to keep something complex (e.g. a form in the dropdown), it would certainly disrupt it. OTOH, is `DropdownPanelView` the right component for such applications?

oleq:
It turned out the `selectstart` event does the same trick but it's less aggressive, e.g. it doesn't break item focus in the dropdown list.

Reinmar:
Couldn't this fix land in the dropdown panel? It'd handle all panels and that's kinda panels' issue because they display over the content. Similar problem may also apply to balloons but I never experienced it.

oleq:
The fix is already in the `dropdownpanelview.js`.

Reinmar:
The patch doesn't fix the issue in Safari.

Reinmar:
And it's this kind of issue which should have a manual test.

oleq:
I created a manual test for the issue.  

As for Safari, it doesn't need the fix at all. Although visually the selection is no longer highlighted in the browser after dropping, press Esc twice to go back to editable and you'll find out that the selection is restored properly.

Reinmar:
Pity that in Safari the selection disappears for a while, but it indeed works fine. We can live with that for some time.

PR #254, opened 2017-06-19
T/248: Made UI destruction a repeatable process

### LINKED ISSUE #248: UI destruction needs to be repeatable
As explained in https://github.com/ckeditor/ckeditor5/issues/114#issuecomment-306522000 and in the following comments.

> Theoretically, it's simple – if you initialise some components in A, B, C order, then they need to be destroyed in C, B, A order.
>
> So, e.g. if the editor first loads plugins and then inits the UI, the UI needs to be destroyed before the plugins are destroyed.
>
> However, what if one of the plugins created some UI and added it to editor UI's collection? On editor destroy, first the UI will be destroyed (which is recursive and will destroy that plugin's component too) and then the plugins will be destroyed and one of these plugins may want to somehow destroy the component which it created (it might've even initialised it, but usually plugins don't do that).
> 
> We've got a collision of responsibilities here. editor.ui.destroy() is recursive so it destroys things which the editor creator didn't create. And then plugins come and try to destroy those things again.

The solution is explained in https://github.com/ckeditor/ckeditor5/issues/114#issuecomment-307730049.

oleq:

Other: Made the UI destruction a fail–safe, repeatable process. Closes #248.

### DIFF
--- a/src/panel/balloon/contextualballoon.js
+++ b/src/panel/balloon/contextualballoon.js
@@ -205,14 +205,4 @@ export default class ContextualBalloon extends Plugin {
 	_getBalloonPosition() {
 		return this._stack.values().next().value.position;
 	}
-
-	/**
-	 * @inheritDoc
-	 */
-	destroy() {
-		this.editor.ui.view.body.remove( this.view );
-		this.view.destroy();
-		this._stack.clear();
-		super.destroy();
-	}
 }
--- a/src/view.js
+++ b/src/view.js
@@ -14,7 +14,6 @@ import DomEmmiterMixin from '@ckeditor/ckeditor5-utils/src/dom/emittermixin';
 import ObservableMixin from '@ckeditor/ckeditor5-utils/src/observablemixin';
 import Collection from '@ckeditor/ckeditor5-utils/src/collection';
 import mix from '@ckeditor/ckeditor5-utils/src/mix';
-import log from '@ckeditor/ckeditor5-utils/src/log';
 import isIterable from '@ckeditor/ckeditor5-utils/src/isiterable';
 
 /**
@@ -101,16 +100,6 @@ export default class View {
 		 * @member {module:ui/viewcollection~ViewCollection}
 		 */
 		this._unboundChildren = this.createCollection();
-
-		/**
-		 * Specifies whether the instance was destroyed using {@link #destroy} method
-		 * in the past.
-		 *
-		 * @private
-		 * @readonly
-		 * @member {Boolean}
-		 */
-		this._wasDestroyed = false;
 
 		// Pass parent locale to its children.
 		this._viewCollections.on( 'add', ( evt, collection ) => {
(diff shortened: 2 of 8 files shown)


oleq:
I wonder if this helps with https://github.com/ckeditor/ckeditor5-core/pull/87. @pomek?

Reinmar:
Why did you add the logging? What's the purpose of it?

The issue would be that someone would try to use a component which was already destroyed. Not that someone attempted to destroy it again.

Let's go easy here. Besides simple KISS rule, we need to be consistent here. If we'd introduce warning here, we should do that in **all** other destroyable classes.

Reinmar:
In other words, let's not make our lives miserable for no good reason.

Reinmar:
All editor tests pass without any problems with these changes.

oleq:
> Why did you add the logging? What's the purpose of it?

I found many bugs this way and I decided it's a good thing to keep it. We wouldn't be here in the first place if not for the problems with the double destruction, right? 

Even if it's now possible to call successive `destroy()` on the same `View`, I still consider it a possible issue in the controller. In other words: it works, but if you do this, your code is probably wrong and will blow up anyway this way or another. Just as you shouldn't call `init()` twice ([we even have an error for this](https://github.com/ckeditor/ckeditor5-ui/blob/master/src/view.js#L290)), you shouldn't `destroy()` stuff many times either. It's a debugging hell lurking in the code, waiting to unleash its true power when you build something more complex on top of it.

And no, we won't have unit tests to check this thing because you'd need to remember about this case in dozens of controllers in the years to come. If that was so easy, we wouldn't be here. We failed to discover the problem for months and I'm afraid it's not the last time we're facing this problem.

Reinmar:
I understand your way of thinking and if that was a cheap approach, I'd vote for it. But it isn't. We must be smart when adding excessive logging because it always means more test, more code to maintain, bigger code size, and so on. 

I was actually complaining in the past about your choices what to validate because you tended to validate things like types of params. Those kinds of issues, with wrong param types in such places like API entry points are very easy to track. We should add warnings in places which are prone to hardly trackable runtime state. Inside complex algorithms, inside complex flows and whenever some important assumption is made which result might come out in a completely different place. Not on the entrance, when things are still easy to follow. Not in a place where 3 lines below an error would be thrown anyway.

And double destroy is such a thing – it's easy to follow **if it causes any bug**. If any day it would cause us a problem, you wouldn't need a day to find it – it'd be rather obvious. Especially that we're talking about destruction which would need to either throw... or I don't know what cause there's no state after destroy.

KISS. 

> I found many bugs this way and I decided it's a good thing to keep it.

So we do with tests. But we don't bundle tests into the code ;)

oleq:
> because it always means more test

Like a single test that verifies if it actually warns?

> more code to maintain

Like a single line and a bunch lines of comment?

>  bigger code size

And yet we use it in [situations](https://github.com/ckeditor/ckeditor5-adapter-ckfinder/blob/master/src/uploadadapter.js#L50) where things obviously blow up too. It's hard to `XMLHttpRequest.open()` without an url, right?

> I was actually complaining in the past about your choices what to validate because you tended to validate things like types of params. Those kinds of issues, with wrong param types in such places like API entry points are very easy to track.

I agree. But let's focus on the present issue. I just find that we have tools (`log.warn`) which may improve DX and yet we hardly use it. Just a couple of warns across thousands LOC. My point is that...

> And double destroy is such a thing – it's easy to follow if it causes any bug.

...it's not only about us. We should support developers who might not be familiar with some concepts and more often than not they develop things blindly. We've been on SO for years now and people struggle even with easy, documented tasks. Helping them out with `warn` here and there is a very low–cost solution IMO. Unlike errors, which must be caught, you can add and remove warnings whenever you think someone may need them.

PR #264, opened 2017-07-04
t/225: Made the UI component initialisation and destruction processes synchronous

### LINKED ISSUE #225: Do we need support for asynchronous components initialisation?
Long, long time ago we assumed that our UI framework needs to support async components. An example of an async component is one which uses an iframe. For it to be useful, ready to work, we need to wait for the iframe's load event. Otherwise, we wouldn't be able to put child components inside this one (if it had some collection).

So, examples of async components are, for instance:

* iframed editable,
* iframed dropdown panel list (used extensively in CKE4, not used at all in CKE5, but may be needed one day),
* some rich UI component which requires lazy loading (big colorpicker widget :D).

In order to support this, we split component initialisation into two parts – the constructor and potentially async `init()` (it can return a promise). We also allowed `destroy()` to be asynchronous (just in case). All this meant that `ViewCollection.add/destroy()` are asynchronous too. (I noticed that `remove()` isn't async... which may be a bug in this plan, cause removing an iframe unloads it, so it's close to destroy :D).

It'd be good if this was where the asynchronous execution would end. But if one of your methods returns a promise then you need to carry it around everywhere and all your code needs to deal with an asynchronous execution. I report this ticket exactly because we noticed that a high level component (`ContextualBalloon`) has a problem with it.

## Do we need this at all?

The first question which we should ask immediately is – do we need all these problems at all? Do we really need to support asynchronous components initialisation?

The idea behind creating our own UI framework was that we have a very limited and quite specific use case so we might be able to come up with something specialised and simple at the same time. Async support works against this. It heavily complicates our code.

At the same time, we don't have any async component yet and, even though we implemented iframed editable in the past to see how the concept works, I'm pretty sure that we missed something (just like the `remove()` issue).

So, perhaps we can drop support for async initialisation which will allow simplifying the code significantly. It will not only affect the UI lib but also features.

But what about those special, rare cases where we really need to deal with asynchronousity?

## Who should be concerned about component's state?

Disclaimer: It may be a total BS what I'll write here because I didn't do any research and I haven't tested this concept, but let's see.

The idea about Web Components is dear to me. It assumes that every component is totally independent which makes it reusable and allows easy composition. Components should have, therefore, a clear and simple API. Just like the entire DOM.

But how does DOM handle asynchronousity? Well, it doesn't (I'm not sure about Web Components, though. @Comandeer?).

If you created an `<iframe>` and added it to the DOM using `appendChild()` it's your duty to postpone any further execution until the `load` event. `appendChild()` doesn't return a promise too. Nothing like that.

This means that the platform is simple (yeah... DOM is so simple!) and the potential complexity is shifted to its users. After all, how often do you deal with iframes? Why making the entire API async if you need to deal with async cases once in a lifetime?

Let's also look on this from a different angle. If we have `ContextualBalloon` which delegates the add/remove API of its internal container, then who really needs to care about the fact that an iframe added to the `ContextualBalloon` is not ready immediately? Certainly not the `ContextualBalloon` because it only cares about `<iframe>` being its child. It's the code which creates that iframe component and uses it. So it's the caller, not the callee.

So, tl;dr here is that we made sure that the entire tree of components and the API needs to be aware of the state of some component that may be added to it. But we should do the exact opposite – save everyone from this knowledge and only make the smallest possible piece of code (usually, some glue code or controllers) aware of it.

oleq:

Other: Made the UI component initialization and destruction processes synchronous. Closes #225.

BREAKING CHANGE: `View#init`, `View#destroy` (also `ViewCollection#init`, `ViewCollection#destroy` and `ViewCollection#add`) no longer return `Promise`. It may require updates in UI components which inherit from `View` and rely on the value returned by these methods.

BREAKING CHANGE: Various UI components switched to synchronous `init()` and `destroy()` (no longer returning `Promise`), which means that features using these components may need some updates to work properly.

---

### Additional information

CI branch is https://github.com/ckeditor/ckeditor5/tree/t/ckeditor5-ui/225, also pointing to related branches in other repositories aligning the API to the changes in this issue. They can't be merged one by one so it's all or nothing merge. CI in https://travis-ci.org/ckeditor/ckeditor5/builds/249943542.

Commit message for related branches:

```
Internal: Updated the usage of UI components which are now driven by synchronous initialization and destruction (see ckeditor/ckeditor5-ui#225).
```

@oskarwrobel: Can you take a look on `ContextualBalloon` and `ContextualToolbar` classes? I made them synchronous but AFAIR because of the asynchronous API the tests were full of hacks. Perhaps you may have some ideas as to how to simplify them even more?

### DIFF
--- a/src/toolbar/contextual/contextualtoolbar.js
+++ b/src/toolbar/contextual/contextualtoolbar.js
@@ -90,7 +90,7 @@ export default class ContextualToolbar extends Plugin {
 		const config = this.editor.config.get( 'contextualToolbar' );
 		const factory = this.editor.ui.componentFactory;
 
-		return this.toolbarView.fillFromConfig( config, factory );
+		this.toolbarView.fillFromConfig( config, factory );
 	}
 
 	/**
@@ -142,7 +142,6 @@ export default class ContextualToolbar extends Plugin {
 	 * Fires {@link #event:beforeShow} event just before displaying the panel.
 	 *
 	 * @protected
-	 * @return {Promise} A promise resolved when the {@link #toolbarView} {@link module:ui/view~View#init} is done.
 	 */
 	_showPanel() {
 		const editingView = this.editor.editing.view;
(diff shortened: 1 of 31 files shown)


Reinmar:
🎉

src/toolbar/contextual/contextualtoolbar.js
```diff
@@ -142,55 +142,46 @@ export default class ContextualToolbar extends Plugin {
...
-					// Add panel to the common editor contextual balloon.
-					this._balloon.add( {
-						view: this.toolbarView,
-						position: this._getBalloonPositionData(),
-						balloonClassName: 'ck-toolbar-container ck-editor-toolbar-container'
-					} )
-				);
+		// If `beforeShow` event is not stopped by any external code then panel will be displayed.
+		this.once( 'beforeShow', () => {
+			if ( isStopped ) {
```
Reinmar: AFAIR, this could've been done using event's cancelling mechanism. Maybe we can change it now?

oskarwrobel: Related ticket https://github.com/ckeditor/ckeditor5-ui/issues/232.

Reinmar: OK, so let's not do that now.

src/view.js
```diff
@@ -290,26 +287,20 @@ export default class View {
...
+
+		// Spread the word that this view is ready!
+		this.ready = true;
 	}
 
 	/**
-	 * Destroys the view instance and child views located in {@link #_viewCollections}.
-	 *
-	 * @returns {Promise} A Promise resolved when the destruction process is finished.
+f	 * Destroys the view instance and child views located in {@link #_viewCollections}.
```
Reinmar: Typoed "f".

Reinmar:
I've scanned through the changes in this repo and they are so simple that if the tests pass I'd risk merging without spending more time.

Reinmar:
Changes in the other repos look good too.

Reinmar:
I'll wait for @oskarwrobel's +1 and merge it.

tests/toolbar/contextual/contextualtoolbar.js
```diff
@@ -320,7 +295,6 @@ describe( 'ContextualToolbar', () => {
 		beforeEach( () => {
 			setData( editor.document, '<paragraph>[bar]</paragraph>' );
 
-			// Methods are stubbed because return internal promise which can't be returned in test.
 			showPanelSpy = sandbox.stub( contextualToolbar, '_showPanel', () => {} );
```
oskarwrobel: `stub` can be changed to `spy`.

oleq:
Are we OK to merge the constellation?

Reinmar:
OK from me.

PR #265, opened 2017-07-05
t/153: List items should handle Enter and Space key press when focused

### LINKED ISSUE #153: List component doesn't handle the Enter and Space keys
feb-02-2017 10-51-47

Enter and space keys on buttons are handled, so should be in this case too.

BTW, there's also that issue that the UI events' callbacks don't focus the editor before executing commands. Perhaps, we should work on all that together.

oleq:

Fix: List items should handle Enter and Space key press when focused. Closes #153.

Requires: ckeditor/ckeditor5-utils#169.

### DIFF
--- a/src/list/listitemview.js
+++ b/src/list/listitemview.js
@@ -114,7 +114,7 @@ export default class ListItemView extends View {
 
 		// Execute on Enter and Space key press.
 		this.keystrokes.set( 'Enter', onKeystrokePress );
-		this.keystrokes.set( [ 32 ], onKeystrokePress );
+		this.keystrokes.set( 'Space', onKeystrokePress );
 	}
 
 	/**
--- a/src/list/listitemview.js
+++ b/src/list/listitemview.js
@@ -9,6 +9,7 @@
 
 import View from '../view';
 import Template from '../template';
+import KeystrokeHandler from '@ckeditor/ckeditor5-utils/src/keystrokehandler';
 
 /**
  * The list item view class.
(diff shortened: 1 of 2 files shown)


oleq:
https://github.com/ckeditor/ckeditor5-ui/pull/265/commits/98629db52f5b3fd7fedba01bd5474a6b3ec8f5d5 implements the whole thing in the `Template`. https://github.com/ckeditor/ckeditor5-ui/pull/265/commits/7deb13b5851499ead7ea91d809cad8f3a7dfb114 uses `KeystrokeHandler`, which is IMO a better choice because it separates us from native DOM API (like `preventDefault`) and will be safer in the future as KH gets improved and secured.

src/list/listitemview.js
```diff
@@ -93,9 +102,34 @@ export default class ListItemView extends View {
...
 		this.element.focus();
 	}
 }
+
+// Returns the Enter and Space keystroke handler for given ListItemView.
+//
+// @private
+// @param {module:ui/list/listitemview~ListItemView} view
+// @returns {Function} A listener for {@link module:utils/keystrokehandler~KeystrokeHandler}.
+function getKeystrokePressCallback( view ) {
```
oskarwrobel: Why not a simple local function inside init() method?
```javascript
const keystrokePressCallback = ( data, cancel ) => {
	this.fire( 'execute' );
	cancel();
};
```

src/list/listitemview.js
```diff
@@ -93,9 +102,34 @@ export default class ListItemView extends View {
...
+	 * @inheritDoc
+	 */
+	init() {
+		const onKeystrokePress = getKeystrokePressCallback( this );
+
+		this.keystrokes.listenTo( this.element );
+
+		// Execute on Enter and Space key press.
+		this.keystrokes.set( 'Enter', onKeystrokePress );
+		this.keystrokes.set( [ 32 ], onKeystrokePress );
```
oskarwrobel: I'm wondering why <kbd>Space</kbd> is not defined in our `keyCodes`.

oleq: Me too TBH. I'll create a PR for this.

PR #271, opened 2017-07-10
T/261 Add support for multiple context elements in `clickOutsideHandler`

### LINKED ISSUE #261: clickOutsideHandler and multiple elements
I want to use `clickOutsideHandler` but pass multiple elements as "inside", so clicking on any of these elements will not be considered as clicking "outside". 

On solution would be to make it possible to pass multiple `contextElement` and check all of them if they contain `target`.

Another would be to pass `evt` and `target` to the `activator` function to make it possible to check target in the activator.

ma2ciek:

Other: Add support for multiple context elements in `clickOutsideHandler`. Closes #261.

### DIFF
--- a/src/bindings/clickoutsidehandler.js
+++ b/src/bindings/clickoutsidehandler.js
@@ -10,7 +10,7 @@
 /* global document */
 
 /**
- * Handles a DOM `click` event outside of specified element and fires an action.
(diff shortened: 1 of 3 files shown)


ma2ciek:
Repositories aligned to this change:
 - https://github.com/ckeditor/ckeditor5-link/tree/ckeditor5-engine/t/261

PR #273, opened 2017-07-11
t/272: Implemented the ContextualBalloon#clear method

### LINKED ISSUE #272: Implement ContextualBalloon#clear method
It's needed by https://github.com/ckeditor/ckeditor5-image/issues/96 to get rid of all view in some situations. `CB#remove` is not so pretty and efficient because it means removing the `#visibleView` in a loop until it's `null`, and each time it modifies the DOM.

oleq:

Other: Implemented the `ContextualBalloon#clear` method. Closes #272.

### DIFF
--- a/src/panel/balloon/contextualballoon.js
+++ b/src/panel/balloon/contextualballoon.js
@@ -142,7 +142,7 @@ export default class ContextualBalloon extends Plugin {
 			this._stack.delete( view );
 
 			// Next we need to check if there is other view in stack to show.
-			const last = Array.from( this._stack.values() ).pop();
+			const last = getLastInStack( this._stack );
 
 			// If it is some other view.
 			if ( last ) {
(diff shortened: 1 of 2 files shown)


src/panel/balloon/contextualballoon.js
```diff
@@ -159,6 +159,20 @@ export default class ContextualBalloon extends Plugin {
 	}
 
 	/**
+	 * Removes all views from the balloon, and also hides it.
+	 */
+	clear() {
```
oskarwrobel: This method is dangerous. When some plugin calls it and clears `CB` stack then rest of the plugins which views were added to the stack won't know that its views are removed.

oskarwrobel: Our approach was that only the specific plugin is able to remove its view.

PR #276, opened 2017-07-17
ContextualToolbar is not displayed when all components inside are disabled.

### LINKED ISSUE #232: Change contextualToolbar#beforeShow to contextualToolbar#show
Once we will get rid of this promise https://github.com/ckeditor/ckeditor5-ui/blob/827edae3522e20220a26c1430c5392cdc18404bd/src/toolbar/contextual/contextualtoolbar.js#L153 we should change the  `contextualToolbar#beforeShow` event to `contextualToolbar#show` end execute the default action on this event with the low priority.

It means that instead of:
```js
contextualToolbar.on( 'beforeShow', ( evt, stop ) => {
     stop();
} );
```
I should be able to do:
```js
contextualToolbar.on( 'show', evt => {
     evt.stop();
} );
```

### LINKED ISSUE #269: ContextualToolbar should be not displayed when all components inside are disabled
This will automatically turn off CT when the editor will be in read-only mode (see read-only requirements https://github.com/ckeditor/ckeditor5/issues/492). And this might be a nice way to disable toolbar for specific elements.

oskarwrobel:

Feature: `ContextualToolbar#beforeShow` event is replaced by a `ContextualToolbar#show` event. `ContextualToolbar` is not displayed when all components inside are disabled. Closes #269. Closes #232.

BREAKING CHANGE: `ContextualToolbar#beforeShow` event is replaced by a `ContextualToolbar#show`.

### DIFF
--- a/src/toolbar/contextual/contextualtoolbar.js
+++ b/src/toolbar/contextual/contextualtoolbar.js
@@ -179,7 +179,7 @@ export default class ContextualToolbar extends Plugin {
 	_show() {
 		// Don not show ContextualToolbar when all components inside are disabled
 		// see https://github.com/ckeditor/ckeditor5-ui/issues/269.
-		if ( Array.from( this.toolbarView.items ).every( item => !item.isEnabled ) ) {
+		if ( Array.from( this.toolbarView.items ).every( item => item.isEnabled !== undefined && !item.isEnabled ) ) {
 			return;
 		}
 
--- a/src/toolbar/contextual/contextualtoolbar.js
+++ b/src/toolbar/contextual/contextualtoolbar.js
@@ -177,6 +177,12 @@ export default class ContextualToolbar extends Plugin {
 	 * @private
 	 */
 	_show() {
+		// Don not show ContextualToolbar when all components inside are disabled
+		// see https://github.com/ckeditor/ckeditor5-ui/issues/269.
+		if ( Array.from( this.toolbarView.items ).every( item => !item.isEnabled ) ) {
+			return;
+		}
+
 		// Update panel position when selection changes (by external document changes) while balloon is opened.
 		this.listenTo( this.editor.editing.view, 'render', () => {
 			this._balloon.updatePosition( this._getBalloonPositionData() );
--- a/src/toolbar/contextual/contextualtoolbar.js
+++ b/src/toolbar/contextual/contextualtoolbar.js
@@ -81,8 +81,8 @@ export default class ContextualToolbar extends Plugin {
 		this._handleFocusChange();
 
 		// The appearance of the ContextualToolbar method is event–driven.
-		// It is possible to stop the #show event and thus prevent the toolbar from showing up.
-		this.on( 'show', () => this._show(), { priority: 'low' } );
+		// It is possible to stop the #show event and this prevent the toolbar from showing up.
+		this.decorate( 'show' );
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


src/toolbar/contextual/contextualtoolbar.js
```diff
@@ -195,6 +171,32 @@ export default class ContextualToolbar extends Plugin {
...
 	/**
+	 * Executes showing toolbar.
+	 * When {@link #event:show} is not stopped then toolbar will be displayed.
+	 *
+	 * @private
+	 */
+	_show() {
+		// Don not show ContextualToolbar when all components inside are disabled
+		// see https://github.com/ckeditor/ckeditor5-ui/issues/269.
+		if ( Array.from( this.toolbarView.items ).every( item => !item.isEnabled ) ) {
```
oleq: I'm not sure this is the right way to check it. As we haven't specified the "`ToolbarItem`" interface yet, there could be some items (that we haven't implemented yet) in the toolbar which don't have `isEnabled` attribute defined but they do work. 

The toolbar will hide then and the user will lose the access to some tools (like a stateless "About" and "Source" buttons in v4). I know we don't have any stateless 

image

I think we should check if `isEnabled` is defined first. Or maybe I'm exaggerating, WDYT? :P

oskarwrobel: Fixed.

oleq: Where?

oskarwrobel: Ups `error: failed to push some refs to 'git@github.com:ckeditor/ckeditor5-ui.git'`

oskarwrobel: Now :)

oleq:
I guess we need a fix in https://github.com/ckeditor/ckeditor5-image/blob/master/src/image.js#L51 too.

src/toolbar/contextual/contextualtoolbar.js
```diff
@@ -79,6 +79,10 @@ export default class ContextualToolbar extends Plugin {
 		// Attach lifecycle actions.
 		this._handleSelectionChange();
 		this._handleFocusChange();
+
+		// The appearance of the ContextualToolbar method is event–driven.
+		// It is possible to stop the #show event and thus prevent the toolbar from showing up.
+		this.on( 'show', () => this._show(), { priority: 'low' } );
```
pjasiun: In fact, we could use decorate here, see https://github.com/ckeditor/ckeditor5-utils/blob/f67aea18988fe58db7ac2f9f0042a938cc96e095/src/observablemixin.js#L252-L309

oleq: A good point. Somehow I've missed implementation of this method and I think it's interesting.

pjasiun: We introduced it recently, for commands refactoring. You can find usage here: https://github.com/ckeditor/ckeditor5-core/blob/ffac7e7ed7c1a06142bcb9157dc4d131652d4efd/src/command.js.

oleq:
I fixed the ckeditor5-image in https://github.com/ckeditor/ckeditor5-image/tree/t/ckeditor5-ui/269. Needs to be merged along this PR.

src/toolbar/contextual/contextualtoolbar.js
oskarwrobel: Done.

PR #280, opened 2017-07-24
t/142: Button tooltip should not look blurry on ldpi screens

### LINKED ISSUE #133: Implement tooltips as views to make :after/:before available to the developers
Moved from https://github.com/ckeditor/ckeditor5-ui-default/issues/150.

### LINKED ISSUE #142: [Chrome] Tooltips are super blurry
image

Compare with:

image

I think that it may be related to non-integer value of `left` position of the tooltip... I don't know how it's positioned to 50% of the width of its parent, but perhaps Chrome isn't rounding this value internally?

oleq:

Fix: Button tooltip should not look blurry on ldpi screens. Closes #142. Closes #133.

BREAKING CHANGE: The `ck-tooltip__main`, `ck-tooltip__arrow` and `ck-tooltip__elements` mixins are no longer available. CSS classes should be used instead.

---

### Additional information

* Requires https://github.com/ckeditor/ckeditor5-theme-lark/pull/104.

### DIFF
--- a/src/button/buttonview.js
+++ b/src/button/buttonview.js
@@ -10,6 +10,7 @@
 import View from '../view';
 import Template from '../template';
 import IconView from '../icon/iconview';
+import TooltipView from '../tooltip/tooltipview';
 
 import { getEnvKeystrokeText } from '@ckeditor/ckeditor5-utils/src/keyboard';
 
@@ -140,6 +141,13 @@ export default class ButtonView extends View {
 		 * @member {module:ui/icon/iconview~IconView} #iconView
 		 */
 
+		/**
+		 * Tooltip of the button view.
+		 *
+		 * @readonly
+		 * @member {module:ui/tooltip/tooltipview~TooltipView} #tooltipView
+		 */
+
 		const bind = this.bindTemplate;
 
 		this.template = new Template( {
@@ -148,16 +156,12 @@ export default class ButtonView extends View {
 			attributes: {
 				class: [
 					'ck-button',
-					bind.if( '_tooltipString', 'ck-tooltip_s' ),
 					bind.to( 'isEnabled', value => value ? 'ck-enabled' : 'ck-disabled' ),
 					bind.if( 'isVisible', 'ck-hidden', value => !value ),
 					bind.to( 'isOn', value => value ? 'ck-on' : 'ck-off' ),
 					bind.if( 'withText', 'ck-button_with-text' )
 				],
 				type: bind.to( 'type', value => value ? value : 'button' ),
-				'data-ck-tooltip': [
-					bind.to( '_tooltipString' )
-				],
 				tabindex: bind.to( 'tabindex' )
 			},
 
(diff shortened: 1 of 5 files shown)


Mgsy:
It seems that this fix solves the problem with blurry tooltips. Text on labels is sharp and you can easily read it. 

Also, I didn't notice any issues caused by this fix.

oskarwrobel:
My only doubt is that tooltip is not so easy to use. I used to use tooltips that can be easily attached to any element by simple API or `[data-*]` attributes.

oskarwrobel:
I'm wondering if tooltip could be a mixin? This mixin could be mixed with any `View` and then tooltip options could be set by `[data-*]` attributes. Does it make any sense?

oleq:
> I'm wondering if tooltip could be a mixin? This mixin could be mixed with any View and then tooltip options could be set by [data-*] attributes. Does it make any sense?

Some example?

oskarwrobel:
The idea with `[data-*]` was stupid. But still...

```javascript
mix( ButtonView, TooltipMixin )

const button = new ButtonView();

button.tooltip = 'tooltip text';
button.tooltipPosition = 's';
```

Is it possible?

oleq:
I don't think so. You need to either
* extend `ButtonView#template` to inject the toolip DOM elements 
* create the TooltipView and put it in an already rendered `ButtonView#element`

I think neither is going to happen with `mix()`. Inheritance yes, but not sure how `mix()` could help.

Reinmar:
The current code looks ok to me. It's not hard to use the tooltip. You just need to create it in your view, add it as a child of it and set its 2 properties. The button view class exposes few things (the position for instance). But if you don't have such a requirement then adding a tooltip should be 3-4 lines of code. Am I right?

oleq:
What we can do, however, is a helper like this:

```js
enableViewTooltip( view, tooltipTextAttribute, tooltipPositionAttribute ) {
	const tooltipView = new TooltipView();

	tooltipView.bind( 'text' ).to( this, tooltipTextAttribute );
	tooltipView.bind( 'position' ).to( this, tooltipPositionAttribute );
	view.element.appendChild( tooltipView.element );

	// Make sure the tooltip will be destroyed along with the view.
	view.addChildren( tooltipView );

	return tooltipView;
}
```

Reinmar:
It could be a static method of `TooltipView`. E.g. `TooltipView.enableIn( view )`. I'd make the attribute names standardised so you don't have to pass them. If you use different ones – create the view manually. In fact, without binding, if values of these two properties could be passed directly to `TooltipView(), that would be **3 lines of code**:

```js
const tooltip = new TooltipView( 'title', 'position' );
this.element.appendChild( tooltip.element );
this.addChildren( tooltip );
```

oskarwrobel:
> In fact, without binding, if values of these two properties could be passed directly to `TooltipView(), that would be **3 lines of code**

That's true. It's not so complex :)

Code looks fine, tooltips works :ok_hand:. I'm merging it.

oskarwrobel:
Wait :)

> if values of these two properties could be passed directly

This is missing :)

oskarwrobel:
Anyway, for me current code is fine.

oskarwrobel:
It's not enough to implement tooltip in JS. It needs to be handled in CSS as well.

```scss
.ck-button:hover {
	@include ck-tooltip_visible();
}
```

PR #284, opened 2017-08-02
ButtonView#tooltip can be added and removed after ButtonView initialization.

### LINKED ISSUE #283: Cannot easily turn off a tooltip once added
How to easily torn off a tooltip once added?

Now when tooltip value is set as `''` empty tooltip appears.

image

oskarwrobel:

Fix: `ButtonView#tooltip` can be added and removed after `ButtonView` initialization. Closes #283.

### DIFF
--- a/src/button/buttonview.js
+++ b/src/button/buttonview.js
@@ -147,17 +147,18 @@ export default class ButtonView extends View {
 		);
 
 		/**
-		 * Icon of the button view.
-		 *
-		 * @readonly
-		 * @member {module:ui/icon/iconview~IconView} #iconView
-		 */
-
-		/**
 		 * Tooltip of the button view.
 		 *
 		 * @readonly
 		 * @member {module:ui/tooltip/tooltipview~TooltipView} #tooltipView
+		 */
+		this.tooltipView = this._createTooltipView();
+
+		/**
+		 * Icon of the button view.
+		 *
+		 * @readonly
+		 * @member {module:ui/icon/iconview~IconView} #iconView
 		 */
 
 		const bind = this.bindTemplate;
@@ -190,7 +191,8 @@ export default class ButtonView extends View {
 							text: bind.to( 'label' )
 						}
 					]
-				}
+				},
+				this.tooltipView
 			],
 
 			on: {
@@ -233,17 +235,6 @@ export default class ButtonView extends View {
 			this.addChildren( iconView );
 		}
 
-		if ( this.tooltip ) {
-			const tooltipView = this.tooltipView = new TooltipView();
-
-			tooltipView.bind( 'text' ).to( this, '_tooltipString' );
-			tooltipView.bind( 'position' ).to( this, 'tooltipPosition' );
-			this.element.appendChild( tooltipView.element );
-
-			// Make sure the tooltip will be destroyed along with the button.
-			this.addChildren( tooltipView );
-		}
-
 		super.init();
 	}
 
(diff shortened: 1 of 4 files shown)


src/tooltip/tooltipview.js
```diff
@@ -58,7 +58,8 @@ export default class TooltipView extends View {
 			attributes: {
 				class: [
 					'ck-tooltip',
-					bind.to( 'position', position => 'ck-tooltip_' + position )
+					bind.to( 'position', position => 'ck-tooltip_' + position ),
+					bind.to( 'text', value => !value ? 'ck-hidden' : '' )
```
oskarwrobel: Should I `trim()` this value?

oleq: I don't think so, it would be an over–engineering.

How about `bind.if()` with negative condition?

src/button/buttonview.js
```diff
@@ -255,6 +246,21 @@ export default class ButtonView extends View {
 	}
 
 	/**
+	 * Creates TooltipView instance and bind with button properties.
+	 *
+	 * @private
+	 * @returns {TooltipView}
```
oleq: This will fail in the API docs builder.

tests/button/buttonview.js
```diff
@@ -151,6 +107,15 @@ describe( 'ButtonView', () => {
 					expect( view.tooltipView.text ).to.equal( 'bar (A)' );
 				} );
 
+				it( 'not render tooltip #tooltip value is false', () => {
```
oleq: I don't get this description. What did you mean?

src/button/buttonview.js
```diff
@@ -277,12 +283,12 @@ export default class ButtonView extends View {
 
 				if ( tooltip instanceof Function ) {
 					return tooltip( label, keystroke );
-				} else if ( tooltip === true ) {
```
oleq: How does it work now? 
* Instance of `String` => a string,
* Instance of `Function` => the result of `function()`,
* `true` => ?
* `false` => ?

oskarwrobel: `true` => `_tooltipString` (if available)
`false` => nothing

oleq: `_tooltipString` is a private attribute. No one can access it from the outside. What I meant was: 
* what to do to disable to the tooltip?
* what to do to display the tooltip in the "label (keystroke)" format?

This should be described in the documentation of the `#tooltip` attibute.

tests/button/buttonview.js
oskarwrobel: Fixed.

src/button/buttonview.js
oskarwrobel: Fixed.

src/button/buttonview.js
oskarwrobel: `''` or `false` value will disbale tooltip.
`true` will display label (keystroke) tooltip.

Isn't it already described here?
```
* * If defined as a `Boolean` (e.g. `true`), then combination of `label` and `keystroke` will be set as a tooltip.
* * If defined as a `String`, tooltip will equal the exact text of that `String`.
* * If defined as a `Function`, `label` and `keystroke` will be passed to that function, which is to return
* a string with the tooltip text.
```

oskarwrobel: Nothing's changed here except we don't show an empty tooltip.

oleq: OK. Because I got confused by this particular change and I thought it works in a different way now. Thanks.

PR #288, opened 2017-08-11
T/260: Make limiter element more configurable

### LINKED ISSUE #260: Make limiter element more configurable
Let's assume that someone has a sticky header over the editable, and doesn't want to show BPV over the header and doesn't want to hide BPV under the header.

image

There is no way to add sticky header height to the limiter top offset.

oleq:

Feature: Allowed `BalloonPanelView` position limiter defined as a function. Made `ContextualBalloon` position limiter configurable via `#positionLimiter` property. Closes #260.

---

### Additional information

The whole solution is based on the assumption that it's up to the editor(ui) creator to define the rightful limiter for all floating panels. So far, in all our creators, all the balloons we use (link, image, contextual toolbar) use the same positioning limiter, so I guess it's the right assumption to start with. It does not hard code anything in the `BalloonPanelView` or `ContextualBalloon`, so developers can still do whatever they want with their balloons.

#### Requires the following branches

Branch constellation/CI in https://github.com/ckeditor/ckeditor5/compare/t/ckeditor5-ui/260.

* https://github.com/ckeditor/ckeditor5-utils/tree/t/ckeditor5-ui/260 – brings the support for the functional limiter in `getOptimalPosition` utility.
* https://github.com/ckeditor/ckeditor5-image/tree/t/ckeditor5-ui/260 - makes the image UI respect the `config.ui.balloonLimiter` config.
* https://github.com/ckeditor/ckeditor5-link/tree/t/ckeditor5-ui/260 - makes the link UI respect the `config.ui.balloonLimiter` config.
* https://github.com/ckeditor/ckeditor5-editor-balloon-toolbar/tree/t/ckeditor5-ui/260 - defines the `config.ui.balloonLimiter` config.
* https://github.com/ckeditor/ckeditor5-editor-classic/tree/t/ckeditor5-ui/260 - defines the `config.ui.balloonLimiter` config.
* https://github.com/ckeditor/ckeditor5-editor-inline/tree/t/ckeditor5-ui/260 - defines the `config.ui.balloonLimiter` config.
* https://github.com/ckeditor/ckeditor5-core/tree/t/ckeditor5-ui/260 - documents the `config.ui`.

#### Issues

* How to configure the `config.ui.balloonLimiter`?  The `ui` namespace could land in https://github.com/ckeditor/ckeditor5-core/blob/master/src/editor/editorconfig.jsdoc. Should `*EditorUI` classes document the `balloonLimiter`?
* I wonder if we ever gonna face a situation that within an editor type (creator) X and Y features will need to use different position limiters? If so, this solution is broken. I was against `config.link.balloonLimiter`, `config.image.balloonLimiter` etc. for the sake of the simplicity, though.
* I totally hate the `selectionRootObtainer` helper. 
   * It does not deserve to become an `editor.editing.view` method.
   * OTOH repeating it in each `*EditorUI` makes no sense.
   * I got no idea where to put it anyway. There's too much engine in it to keep it in the `ckeditor5-ui`. OTOH, it does not fit in core, etc.
   * Maybe we can get rid of it somehow?


Reinmar:
> How to configure the config.ui.balloonLimiter? The ui namespace could land in https://github.com/ckeditor/ckeditor5-core/blob/master/src/editor/editorconfig.jsdoc. Should *EditorUI classes document the balloonLimiter?

The UI namespace should be documented in ckeditor5-ui. Can be a jsdoc file there defining similar interfaces like HeadingConfig and ImageConfig.

oleq:
More doubts:
* Since all editor types have only one editable so far, do we need `selectionRootObtainer` at all? We could define the `config.ui.balloonLimiter` as a static HTML element instead.
* I [documented](https://github.com/ckeditor/ckeditor5-core/commit/9b2d0a9452e67a5a6db241f4852d02219f76dee4) the `config.ui.balloonLimiter` and, TBH, it feels more like a property (of what?) than actual config. I'm not sure if developers will actually use it and it's totally internal if you read the docs.

oskarwrobel:
Maybe this limiter should be `EditorUi` getter `EditorUI#balloonLimiter`? EditorUI class knows the editor.

oskarwrobel:
And maybe this is the right time for common `EditorUi` class? We have `ClassidEditorUi`, `InlineEditorUI` and `BalloonToolbarEditorUI`. All have common stuff like
```
		/**
		 * @inheritDoc
		 */
		this.editor = editor;

		/**
		 * @inheritDoc
		 */
		this.view = view;

		/**
		 * @inheritDoc
		 */
		this.componentFactory = new ComponentFactory( editor );

		/**
		 * @inheritDoc
		 */
		this.focusTracker = new FocusTracker();
```

Reinmar:
> All have common stuff like

We had `EditorUI` in the past but removed it in favour of an interface. We plan to do the same with `Editor` classes for the same reasons – it's hard to define a one-to-rule-them-all base class.

oskarwrobel:
Makes sense. But still, `ballloonLimiter` fits fine as EditorUi getter to me.

oskarwrobel:
I've tested this config option and works 👌.

Reinmar:
We discussed where to expose the option to control this and we decided to put it as a `ContextualBalloon` plugin's property. Concrete editor UI implementations or editor integrations will be able to override it. Let's not expose it prematurely in the config as it indeed is pretty hardcore.

Reinmar:
PS. Adding the property to `ContextualBalloon` will prevent from adding it all around the code in the features.

oleq:
I updated the solution and used `ContextualBalloon#positionLimiter` instead of the config, which notably simplified the code. 

Branch constellation/CI in https://github.com/ckeditor/ckeditor5/compare/t/ckeditor5-ui/260.

* https://github.com/ckeditor/ckeditor5-utils/tree/t/ckeditor5-ui/260 – brings the support for the functional limiter in `getOptimalPosition` utility.
* https://github.com/ckeditor/ckeditor5-image/tree/t/ckeditor5-ui/260 - makes the image UI use the default `ContextualBalloon` limiter.
* https://github.com/ckeditor/ckeditor5-link/tree/t/ckeditor5-ui/260 - makes the link UI use the default `ContextualBalloon` limiter.

@oskarwrobel: can you give me the feedback?

oskarwrobel:
Works perfectly.

src/panel/balloon/contextualballoon.js
```diff
@@ -48,6 +49,28 @@ export default class ContextualBalloon extends Plugin {
 		this.view = new BalloonPanelView();
 
 		/**
+		 * The {@link module:utils/dom/position~Options#limiter position limiter}
+		 * for the {@link view}, used when no `limiter` has been passed into {@link #add}
```
oskarwrobel: image

oleq: Fixed in https://github.com/ckeditor/ckeditor5-ui/pull/288/commits/13ae5be5b1d3cc9b3e427e0545f0b0369077db47.

PR #290, opened 2017-08-22
T/289: Changed type of collection used in "Template" class.

### LINKED ISSUE #289: InputTextView can not be added as a child when has no id property or id is a number
When `InputTextView` with no id is defined as a child in template then exception is thrown:
```
CKEditorError: collection-add-invalid-id
```
This is most probably because of missing custom `idProperty`:
https://github.com/ckeditor/ckeditor5-ui/blob/b2d9002f5a829cf278c3258cba8336230f04708f/src/template.js#L1084

https://github.com/ckeditor/ckeditor5-ui/blob/b2d9002f5a829cf278c3258cba8336230f04708f/src/viewcollection.js#L29-L31

pomek:

Fix: The `Template` class will not throw an error if its child does not have an `idProperty`. Closes #289.

### DIFF
--- a/tests/template.js
+++ b/tests/template.js
@@ -571,15 +571,15 @@ describe( 'Template', () => {
 				);
 			} );
 
-			// #289
+			// https://github.com/ckeditor/ckeditor5-ui/issues/289
 			it( 'does not throw when child does not have an "id" property', () => {
+				const strongView = getView( {
+					tag: 'strong'
+				} );
+
+				strongView.set( 'id' );
+
 				expect( () => {
-					const strongView = getView( {
-						tag: 'strong'
-					} );
-
-					strongView.set( 'id' );
-
 					getView( {
 						tag: 'div',
 						children: [
--- a/tests/template.js
+++ b/tests/template.js
@@ -9,7 +9,6 @@ import testUtils from '@ckeditor/ckeditor5-core/tests/_utils/utils';
 import { default as Template, TemplateToBinding, TemplateIfBinding } from '../src/template';
 import View from '../src/view';
 import ViewCollection from '../src/viewcollection';
-import InputTextView from '../src/inputtext/inputtextview';
 import Model from '../src/model';
 import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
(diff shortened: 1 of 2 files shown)


tests/template.js
```diff
@@ -121,6 +122,23 @@ describe( 'Template', () => {
 			expect( tpl.attributes.a[ 0 ] ).to.equal( 'foo' );
 			expect( tpl.children.get( 0 ).tag ).to.equal( 'span' );
 		} );
+
+		it( 'does not throw when child does not have an "id" property', () => {
```
oleq: This test should be in the ["children"](https://github.com/ckeditor/ckeditor5-ui/blob/master/tests/template.js#L389) section.

tests/template.js
```diff
@@ -121,6 +122,23 @@ describe( 'Template', () => {
...
+			class DivView extends View {
```
oleq: You don't have to create a new class to test this. It doesn't bring anything good to the test case but makes it more complex to understand. Use the `getView()` helper instead.

tests/template.js
```diff
@@ -121,6 +122,23 @@ describe( 'Template', () => {
...
+				constructor( locale ) {
+					super( locale );
+
+					this.template = new Template( {
+						tag: 'div',
+						children: [
+							new InputTextView( locale )
```
oleq: You shouldn't use the `InputTextView` either. This thing should work for any generic view which has `view.set( 'id' )` without a value. 

Besides, if the API of the `InputTextView` changes in the future, this test may become obsolete/unstable. What if `InputTextView#id` is removed? This case wouldn't test anything, it would pass despite regressions in the `Template` class.

In the framework class' tests, we must make sure the cases are generic. We can't rely on any component, which is built on top of such a class.

tests/template.js
```diff
@@ -121,6 +122,23 @@ describe( 'Template', () => {
...
+						]
+					} );
+				}
+			}
+
+			expect( () => new DivView( {} ) ).to.not.throw( CKEditorError );
```
oleq: Simply `to.not.throw()`. There's no point in narrowing this assertion.

pomek:
@oleq, thanks for the tips. I fixed the test. Could you look at this once again?

PR #292, opened 2017-08-23
t/291: ToolbarView#fillFromConfig should warn when the factory does not provide a component

### LINKED ISSUE #291: The ToolbarView should warn when the factory does not provide a component
A follow-up of https://github.com/ckeditor/ckeditor5/issues/526.

oleq:

Other: ToolbarView#fillFromConfig should warn when the factory does not provide a component. Closes #291. Closes https://github.com/ckeditor/ckeditor5/issues/526.

---

### Additional information

* Could close https://github.com/ckeditor/ckeditor5/issues/526.

### DIFF
--- a/src/toolbar/toolbarview.js
+++ b/src/toolbar/toolbarview.js
@@ -140,8 +140,7 @@ export default class ToolbarView extends View {
 				 * to provide a toolbar item has not been loaded or there is a typo in the configuration.
 				 *
 				 * Make sure the plugin responsible for this toolbar item is loaded and the toolbar configuration
-				 * configuration is correct, e.g. {@link module:basic-styles/bold~Bold} is loaded for the `'bold'`
-				 * toolbar item.
+				 * is correct, e.g. {@link module:basic-styles/bold~Bold} is loaded for the `'bold'` toolbar item.
 				 *
 				 * @error toolbarview-item-unavailable
 				 * @param {String} name The name of the component.
--- a/src/toolbar/toolbarview.js
+++ b/src/toolbar/toolbarview.js
@@ -133,20 +133,22 @@ export default class ToolbarView extends View {
 				this.items.add( factory.create( name ) );
 			} else {
 				/**
-				 * There was a problem with expanding the toolbar configuration into toolbar items.
-				 * The provided factory does not provide a component of such a name and because of that
-				 * it has not been added to the {@link #items}.
+				 * There was a problem processing the configuration of the toolbar. The item with the given
+				 * name does not exist and it was omitted when adding toolbar items in DOM.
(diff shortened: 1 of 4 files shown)


src/toolbar/toolbarview.js
```diff
@@ -126,9 +127,29 @@ export default class ToolbarView extends View {
...
+				 * This warning usually shows up when the plugin that is supposed to register the toolbar
+				 * component in the factory has not been loaded or there's a typo in the configuration
+				 * of the toolbar.
+				 *
+				 * @error toolbarview-missing-component
+				 * @param {String} name The name of the component.
+				 * @param {module:ui/componentfactory~ComponentFactory} factory The factory that is missing the component.
+				 */
+				log.warn(
+					'toolbarview-missing-component: There is no such component in the factory.',
```
Reinmar: This was supposed to be a more meaningful error, but it's still all about esoteric components and factories. I think you should read https://medium.com/@thomasfuchs/how-to-write-an-error-message-883718173322 ;)

src/toolbar/toolbarview.js
```diff
@@ -126,9 +127,29 @@ export default class ToolbarView extends View {
...
-			const component = name == '|' ? new ToolbarSeparatorView() : factory.create( name );
-
-			this.items.add( component );
+			if ( name == '|' ) {
+				this.items.add( new ToolbarSeparatorView() );
+			} else if ( factory.has( name ) ) {
+				this.items.add( factory.create( name ) );
+			} else {
+				/**
+				 * There was a problem with expanding the toolbar configuration into toolbar items.
```
Reinmar: Same here as below – "expanding" – really? There should be a registry of forbidden words. Here, try not to use "expanding" and "factory" and "components" :P

Reinmar:
R- because of the wording (which is the essential part o this ticket).

oleq:
Hopefully, this will look more understandable now.

Reinmar:
Since there's only one factory of UI items in the editor I'll remove it from the error details. This will make reading the error easier since there will be only the critical information.

oleq:
> Since there's only one factory of UI items in the editor I'll remove it from the error details. This will make reading the error easier since there will be only the critical information.

Don't forget that ckeditor5-ui is not only about CKEditor, it's a UI framework and it could be used anywhere, each time with a different factory.

Reinmar:
> Don't forget that ckeditor5-ui is not only about CKEditor, it's a UI framework and it could be used anywhere, each time with a different factory.

Sure. And this will happen how many times? ;)

Besides, in such special use cases, the developer will be more aware of what they are doing anyway.

PR #295, opened 2017-09-04
Fix: Removed hacks from `BalloonPanelView` that made BalloonPanel unselectable

### LINKED ISSUE #294: Content inside BalloonPanelView is not selectable
This is probably OK when BPV is used for showing form but when you want to show e.g. notifications it should be possible to select and copy text.

This is probably the reason why BPV is unselectable https://github.com/ckeditor/ckeditor5-ui/issues/243.

oskarwrobel:

Fix: Removed hacks from `BalloonPanelView` that made BalloonPanel content unselectable. Closes #294. Closes https://github.com/ckeditor/ckeditor5/issues/498.

---

### Additional information

@Mgsy could confirm this PR does not make a regression especially for this issues:
- https://github.com/ckeditor/ckeditor5-link/issues/126
- https://github.com/ckeditor/ckeditor5-link/issues/90

### DIFF
--- a/src/panel/balloon/balloonpanelview.js
+++ b/src/panel/balloon/balloonpanelview.js
@@ -13,7 +13,6 @@ import { getOptimalPosition } from '@ckeditor/ckeditor5-utils/src/dom/position';
 import isRange from '@ckeditor/ckeditor5-utils/src/dom/isrange';
 import isElement from '@ckeditor/ckeditor5-utils/src/lib/lodash/isElement';
 import toUnit from '@ckeditor/ckeditor5-utils/src/dom/tounit';
-import preventDefault from '../../bindings/preventdefault.js';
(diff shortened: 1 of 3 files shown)


oleq:
Even if this PR does not cause any regression, we must understand why first. What mechanism made those fixes obsolete?

oskarwrobel:
I think this is obsolete because now `LinkFormView` is focusable and this is not a problem anymore (it was before) because we are using `FocusCycler`.

oleq:
Is the Image form also focusable?

Mgsy:
@oskarwrobel: I can confirm that this PR doesn't make any regression. It also works great and fixes https://github.com/ckeditor/ckeditor5/issues/498.

oleq:
Thanks, @Mgsy. Let's give it a try then.

PR #300, opened 2017-09-12
Changed StickyToolbarView to StickyPanelView

### LINKED ISSUE #297: Create StickyPanelView
We already have a `StickyToolbarView` that extends `ToolbarView`. I'm thinking about creating a separate view instead, like `BPV` with content collection and add `ToolbarView` as a child. This will make possible to easily create any sticky bar not only sticky toolbar.

oskarwrobel:

Feature: Changed `StickyToolbarView` to `StickyPanelView`.  Closes #297.

---

### Related:
- https://github.com/ckeditor/ckeditor5-theme-lark/pull/109
- https://github.com/ckeditor/ckeditor5-editor-classic/pull/64

### For CI:
 - https://github.com/ckeditor/ckeditor5/tree/t/ckeditor5-ui/297


oskarwrobel:
I'm wondering how to disable placeholder element that is automatically injected by a `StickyPanelView` to prevent flickering of the UI. This placeholder is OK when panel has a `position: static` initially but when panel is floating from the beginning then placeholder should be not injected.

oleq:
I just noticed that `BalloonPanelView` has a default `1px` border, while `StickyPanelView` not. I think it should be consistent.

oskarwrobel:
> I just noticed that BalloonPanelView has a default 1px border, while StickyPanelView not. I think it should be consistent.

Will be done here https://github.com/ckeditor/ckeditor5-theme-lark/issues/110.

PR #304, opened 2017-09-19
t/303: Implemented View#removeChildren

### LINKED ISSUE #303: How to remove child view added by addChildren()
It is possible to add a child view omitting `ViewCollection`. We do it eg in `ButtonView`
https://github.com/ckeditor/ckeditor5-ui/blob/606045953783b4ba5f90d7bac885e6efdb568b18/src/button/buttonview.js#L229-L235

`addChildren()` calls init on new view and register this view as a child thanks to this view will be destroyed together with its parent.

But there is no way to remove such a view before the parent will be removed. Something as `removeChildren()` is needed.

oleq:

Feature: Implemented `View#removeChildren`, the opposite of `View#addChildren`. Closes #303.

### DIFF
--- a/src/view.js
+++ b/src/view.js
@@ -275,6 +275,22 @@ export default class View {
 	}
 
 	/**
+	 * The opposite of {@link #addChildren}. Removes a child view from this view instance.
+	 * Once removed, the child is no longer managed by its parent, e.g. it can be safely used elsewhere,
+	 * becoming a child of another parent view.
+	 *
+	 * @see #addChildren
+	 * @param {module:ui/view~View|Iterable.<module:ui/view~View>} children Children views to be registered.
+	 */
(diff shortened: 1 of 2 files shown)


src/view.js
```diff
@@ -275,6 +275,22 @@ export default class View {
 	}
 
 	/**
+	 * The opposite of {@link #addChildren}. Removes a child view from this view instance.
+	 * Once removed, the child is no longer managed by its parent, e.g. it can be safely used elsewhere,
+	 * becoming a child of another parent view.
+	 *
+	 * @see #addChildren
+	 * @param {module:ui/view~View|Iterable.<module:ui/view~View>} children Children views to be registered.
```
oskarwrobel: `Children views to be removed.`

Issue #262, opened 2017-07-03
Should first call of view.element render view.template?


oskarwrobel:
This might cause confusing situations.

Using `this.element` inside view constructor (eg to initialize `FocusTracker` or `KeystrokeHandler`) renders the template and makes it not possible to be extended.
```javascript
class MyView extends View {
    constructor( locale ) {
        super( locale );
         
        this.template = new Template( ... );

        this.focusTracker = new FocusTracker();
        this.fousTracker.add( this.element );
    }
}

const view = new MyView( locale );

Template.extend( view.template, { .... } );
```
And I see a warning:
```
template-extend-render: Attempting to extend a template which has already been rendered.
```
but my view is not even initialized.

To workaround this problem I can move `FT` init to the view `init()` but it took me a while to figure out why I can't extend this template.

oleq:
The very first idea was that `constructor` defines but it's `init` does things related to rendering. Also, if not referenced, the `view#element` does not render, which supposed to be a kind of optimization against unnecessary execution and DOM manipulation.

When we first designed such approach, it looked right. TBH, today, I'm no so sure and it looks to me more like a premature optimization instead. It's confusing in cases like this one. Developers will struggle to understand the magic of `view#element` and it might not be the best DX. 

What if we moved `view.element = this.template.render()` to `View#init`? It would mean that each class inheriting from `View` and implementing own `init` must first call `super.init()` in the method definition. I wonder if that would break something major.

WDYT? (cc @Reinmar)

Reinmar:
There reason why we don't render the `element` in `constructor()` is to allow subclass existing view to change its template. That's the reason why `this.element` should not be accessed in the `constructor()`. Optimisation had nothing to do with that.

I completely agree that this is not the cleanest solution. Perhaps we could replace `init()` with `render()` so it would be clearer that until `render()` is called `this.element` is null and accessing it changes nothing. Less magic, more clarity. WDYT?

oleq:
> There reason why we don't render the element in constructor() is to allow subclass existing view to change its template. That's the reason why this.element should not be accessed in the constructor(). Optimisation had nothing to do with that.

I know, but instead of saying: 

– Don't use `#element` in `constructor` if you want to subclass the `#template`.

we can simply say 

– `view#element` is always null in `constructor` but it is rendered and accessible in `init`.

We don't need to rename `init` to `render` IMO, it doesn't change much in DX.

Reinmar:
[...]

PR #312, opened 2017-09-28
t/311: createListDropdown helper should use clickOutsideHandler

### LINKED ISSUE #311: createListDropdown helper should use clickOutsideHandler
ATM `createListDropdown` uses own custom click handling. It's redundant.

oleq:

Other: The `createListDropdown` helper should use `clickOutsideHandler`. Closes #311.


oleq:
self.r-: This change breaks this test 
https://github.com/ckeditor/ckeditor5-heading/blob/888f9571528f78720c231c19b2358fd5e02f7397/tests/heading.js#L93-L96 as the clickOutsideHandler references `view#element` before `Template.extend` in https://github.com/ckeditor/ckeditor5-heading/blob/master/src/heading.js#L98-L106.

Again, it's all about https://github.com/ckeditor/ckeditor5-ui/issues/262.

oleq:
It turned out this code still won't work. 

It's because the `clickOutsideHandler` [requires a reference to `View#element` in its `contextElements`](https://github.com/ckeditor/ckeditor5-ui/pull/312/files#diff-87ab2335c9320fb35ff63d466896a203R78), which is not available in the `createListDropdown` helper because it returns an instance which is not rendered (yet). And it makes sense, because one might want to customize the instance even further so we cannot simply `render()` it in `createListDropdown`.

Possible solutions:
1. Implement `View#render` event fired after the view is rendered. Then: 
	```js
	function createListDropdown( ... ) {
		const dropdownView = createDropdown( ... );

		// ...

		dropdownView.on( 'render', () => {
			clickOutsideHandler( {
				emitter: dropdownView,
				activator: () => dropdownView.isOpen,
				callback: () => {
					dropdownView.isOpen = false;
				},
				contextElements: [ dropdownView.element ]
			} );
		} );

		// ...
	}
	```    
2. Allow `contextElements` defined as function, so it returns the value (`dropdownView.element`) on demand when the actual `mousedown` is fired and, obviously, the `dropdownView.element` already exists.

Reinmar:
The `#render` event sound fine, as something you'd want anyway in some other place.

Reinmar:
Actually, sorry – `#rendered` because it should be fired once the element is already available.

oleq:
I decorated `View#render` method instead. Seems like the easiest solution.

oleq:
Due to recent changes in the drop-down API, this PR became obsolete. Replaced by https://github.com/ckeditor/ckeditor5-ui/pull/353.

PR #318, opened 2017-10-02
Adding z-index to sticky panel when position: fixed is applied

### LINKED ISSUE #315: Image is rendered above the toolbar
There's some problem with z-indexes again when the `ImageUpload` feature is enabled:

image

szymonkups:

Fix: Fixed sticky panel positioning when above images. Closes #315.

---

### Additional information

Sticky panel which holds the toolbar have `position:fixed` applied, so the elements inside editor with `position:relative` (images for example) are positioned above the sticky panel. Browser compares `z-index` of the sticky panel (which is not set) instead of panel's `z-index: 999`.

### DIFF
--- a/theme/components/panel/stickypanel.scss
+++ b/theme/components/panel/stickypanel.scss
@@ -4,7 +4,7 @@
 @include ck-editor {
 	.ck-sticky-panel {
 		&.ck-sticky-panel_sticky {
-			z-index: ck-z( 'modal' );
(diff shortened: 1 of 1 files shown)


Reinmar:
You could write under the ticket the things you wrote me on priv about why is this needed. For the future ref.

Reinmar:
LGTM.

PR #325, opened 2017-10-16
Enhancement: Make ComponentFactory case-insensitive.

### LINKED ISSUE #324: Make ComponentFactory case-insensitive
http://localhost/ckeditor5/build/docs/ckeditor5/1.0.0-alpha.1/framework/guides/support/error-codes.html#toolbarview-item-unavailable is the most frequently visited error code which means that a lot of people have problems with configuring their toolbars.

It's hard to guess how many of them make mistakes with letter cases but taken that plugins are named with UpperCase and buttons with lowerCase, there's a high chance that they actually do this mistake.

jodator:

Enhancement: Make `ComponentFactory` case-insensitive. Closes #324.

Defining toolbar components should not fail if someone use different case than defined by pluign.

---

### Additional information

* I've added internal method that ensures `name` is a string before executing `.toLowerCase()` just to not have `trying to call .toLowerCase() of undefined` in console.

### DIFF
--- a/src/componentfactory.js
+++ b/src/componentfactory.js
@@ -14,6 +14,7 @@ import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
  *
  * It allows functions producing specific UI components to be registered under their unique names
  * in the factory. A registered component can be then instantiated by providing its name.
+ * Note that names are case-insensitive.
  *
  *		// Editor provides localization tools for the factory.
  *		const factory = new ComponentFactory( editor );
@@ -23,6 +24,9 @@ import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
  *
  *		// An instance of FooView.
  *		const fooInstance = factory.create( 'foo' );
+ *
+ *		// Names are case-insensitive os this is also allowed:
+ *		const barInstance = factory.create( 'Bar' );
  *
  * The {@link module:core/editor/editor~Editor#locale editor locale} is passed to the factory
  * function when {@link module:ui/componentfactory~ComponentFactory#create} is called.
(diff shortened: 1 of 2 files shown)


src/componentfactory.js
```diff
@@ -121,6 +121,18 @@ export default class ComponentFactory {
...
+
+/**
+ * Ensures that component name used as key in internal map is in lower case.
+ *
+ * @ignore
+ * @private
+ * @param {String} name
+ * @returns {String}
+ */
+function _componentsMapKey( name ) {
```
oleq: I'd simply call it `getNormalized( name )`.

src/componentfactory.js
```diff
@@ -121,6 +121,18 @@ export default class ComponentFactory {
 	 * @returns {Boolean}
 	 */
 	has( name ) {
-		return this._components.has( name );
+		return this._components.has( _componentsMapKey( name ) );
 	}
 }
+
+/**
```
oleq: [Private non–class functions](https://github.com/ckeditor/ckeditor5-utils/blob/master/src/emittermixin.js#L559-L579) are documented with `//` in the project. No need to `@ignore` it.

jodator: I saw two patterns for that. I've used the one looked nicer for me ;) But I guess the other pattern is a standard now.

oleq:
A documentation explaining that component names are case insensitive would be nice too. At least in `class ComponentFactory`.

oleq:
@jodator, make sure the merge commit message you propose is compatible with the [convention](https://github.com/ckeditor/ckeditor5-design/wiki/Git-commit-message-convention#convention). Changelog is generated automatically based on this commit schema.

jodator:
@oleq Thanks for reminder. I'm so used to `Enhancement` that I've forgot that it is not official.

PR #352, opened 2017-12-11
t/350: It should be possible to use "dirty" third-party SVGs as the content of the IconView

### LINKED ISSUE #350: Poor SVG icon support in a ButtonView/IconView
I created a toolbar button. When trying to add an icon to it I found that it's really cumbersome. Whatever SVG I tried, it did not render correctly inside the toolbar.
I would expect to see there an SVG image scaled proportionally to the amount of space that is available for it. Instead, I can see just the small top left corner of it (most of the SVGs that I tried had size like 500x500).

The source of the issue lays in the `IconView`, which forces the following:
```
this.set( 'viewBox', '0 0 20 20' );
```

In order to fix my icon I had to set
```
button.iconView.set( 'viewBox', '0 0 128 128' );
```

oleq:

Fix: It should be possible to use "dirty" third-party SVGs as the content of the IconView. Closes #350.

### DIFF
--- a/src/icon/iconview.js
+++ b/src/icon/iconview.js
@@ -75,9 +75,13 @@ export default class IconView extends View {
 	 */
 	_updateXMLContent() {
 		if ( this.content ) {
-			const svg = new DOMParser()
-				.parseFromString( this.content.trim(), 'image/svg+xml' )
-				.firstChild;
+			const parsed = new DOMParser().parseFromString( this.content.trim(), 'image/svg+xml' );
+			const svg = parsed.querySelector( 'svg' );
+			const viewBox = svg.getAttribute( 'viewBox' );
+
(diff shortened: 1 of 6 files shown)


Reinmar:
Can we have a manual test which proves that this PR fixes the issue?

oleq:
Extended existing test in https://github.com/ckeditor/ckeditor5-ui/pull/352/commits/f6f893fac09b0b4f1895a0a4adc2284710aeeccc. Can you check this out, @Reinmar?

Reinmar:
It certainly makes the manual test to pass. But it breaks CI: https://travis-ci.org/ckeditor/ckeditor5-ui/builds/315332565?utm_source=github_status&utm_medium=notification

oleq:
It looks OK now.

PR #359, opened 2018-01-12
T/ckeditor5 font/4

### LINKED ISSUE #358: Change which classes determines dropdown button visual styles
It might be changed by #356 but I think it keeps up with a direction of changes in that ticket.

Right now every "dropdown button" in toolbar gets border and bakcground (like the heading dropdown) and adding type of dropdowns required to add exceptions to that CSS rules in theme.

I think that's better to not add those visual styles to dropdown buttons but instead to only style those "withText" (as in heading dropdown case).

jodator:

Other: Add `ck-dropdown__withtext` class binding to dropdown. Closes #358.

---

### Additional information

* It also requires PR in theme-lark: https://github.com/ckeditor/ckeditor5-theme-lark/pull/121

### DIFF
--- a/src/dropdown/createdropdown.js
+++ b/src/dropdown/createdropdown.js
@@ -44,7 +44,7 @@ export default function createDropdown( model, locale ) {
 	const panelView = new DropdownPanelView( locale );
 	const dropdownView = new DropdownView( locale, buttonView, panelView );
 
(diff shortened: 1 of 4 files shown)


Reinmar:
Can we delete the branch now?

PR #361, opened 2018-01-24
t/344: Refactoring of the dropdown ecosystem

### LINKED ISSUE #344: Implement Split Button dropdown
This UI Component should look like option 2 from [dropdown button reasearch](https://github.com/ckeditor/ckeditor5-ui/issues/333#issuecomment-342806422):

1. Toolbar component should have two separate active parts:
    - action button - that should invoke active action
    - arrow button - that should open dropdown panel
2. The keyboard/mouse actions should act:
    - click on button executes command
    - click on arrow opens dropdown
    - `enter` on button executes command
    - `enter` or `arrow up/dwon` on arrow opens dropdown
3. It should be extensible and allows to implement different UI panels inside dropdown. But the the first implementation should allow to implement [highlight UI](https://github.com/ckeditor/ckeditor5-highlight/issues/3). As such it should expose a way to:
   - change active command/icon and represent it's state (active/enabled).

### LINKED ISSUE #356: Dropdown family (possibly) needs architecture refactoring
ATM there are 2 base classes:
```
DropdownPanelView
DropdownView
```

and 3 helpers combining them with `ToolbarView`, `ListView`, etc.:
```
createDropdown()
createButtonDropdown()
createListDropdown()
```

However, the new [Highlight](https://github.com/ckeditor/ckeditor5-highlight/issues/3) feature we're implementing requires a [split button](https://github.com/ckeditor/ckeditor5-ui/issues/344), 
which does not actually fit that well into the family because it actually comes with 2 buttons in the toolbar:
* main one which applies the highlight, 
* the side one which opens the dropdown with styles.

**The goal**: We need to somehow make it possible to create such a dropdown supporting all events like `execute` and attributes like `isOpen` using just base classes and helper functions. 

## JS requirements

ATM `DropdownView` accepts button and panel as constructor arguments. Then it [extends button's template](https://github.com/ckeditor/ckeditor5-ui/blob/master/src/dropdown/dropdownview.js#L127-L133) and starts [reacting to `ButtonView#execute` by opening the panel](https://github.com/ckeditor/ckeditor5-ui/blob/master/src/dropdown/dropdownview.js#L143-L145).

In case of the split button, the situation is quite different. The left (main) part of the split button usually executes some commands (e.g. applies last used highlight), while the right button acts as the panel opener. So, in fact, the main button is "extra" here, something new.

With the current `DropdownView` architecture, it is not possible to pass such split–button and handle it without lots of hacking.

## CSS requirements

There are 2 dropdown styles in Lark:
1. dropdown with text which has background and border,
2. dropdown without text (just an icon), borderless.

The later is ATM supported only by `createButtonDropdown()` but it should become a standard feature of the `DropdownView`, I guess. Every dropdown should come with that option to use just an icon or just a text, which obviously should be reflected in a class of the element:

```html
<!-- e.g. for headings feature -->
<div class"ck-dropdown ck-dropdown_with-text"></div>
```
vs.
```html
<!-- e.g. for alignment feature -->
<div class"ck-dropdown"></div>
```

The class corresponds to `ck-button_with-text`. 

**Note:** While the styling could change soon and all dropdowns [might look the same](https://github.com/ckeditor/ckeditor5/issues/645), this still must be differentiated.

### Allow passing the split button directly into `DropdownView` constructor

That could work if we figured an interface for such split button which makes sense. Note that both parts of the split button have `isEnabled` attribute and they both fire `execute` when clicked. 

1. [How to allow the `DropdownView` to listen to the "right" `execute` and open the panel?](https://github.com/ckeditor/ckeditor5-ui/blob/master/src/dropdown/dropdownview.js#L143-L145)

2. [How to let the `DropdownView` knew which of the two split buttons is enabled to activate the keyboard support?](https://github.com/ckeditor/ckeditor5-ui/blob/master/src/dropdown/dropdownview.js#L167-L170)

Maybe it should be `createDropdown()'s` job, limiting the `DropdownView` to just being a dull container?

### Let's just create a new kind of `DropdownView` for the split button purposes

It's quite simple but it will duplicate tons of code (and tests) doing exactly the same just to allow a fancy button. It does not make much sense to me, IMO the new feature should remain in the base `DropdownView` family.

## Post-refactoring

Some naming issues should probably be resolved, like this one because we should name dropdown helpers by the kind of content it contains:

```
createButtonDropdown() -> createToolbarDropdown()
```

Which name to use for the "split dropdown" then? `createSplitButtonDropdown`? Probably yes, because it should be able to contain any sort of content. Or maybe we should figure out the new set of helpers according to the following methodology?

```
button type (single vs. split) 
+ 
content type (list vs. toolbar vs. empty) 
= 
result dropdown
```
like
```
create[Split|][List|Toolbar|]Dropdown()
```

I'm not sure how to resolve that. Which helpers should exist and to what extent things should be "helperized"?

cc @Reinmar @jodator @oskarwrobel

jodator:

Enhancement: Introduce `SplitButtonView` and new dropdowns helper methods. Closes #344. Closes #356.

BREAKING CHANGE: Removed `DropdownModel` interface - use dropdown properties directly. See #356.
BREAKING CHANGE: Removed `createButtonDropdown()` and `ButtonDropdownView`. See #356.
BREAKING CHANGE: Removed `createListDropdown()` and `ListDropdownView`. See #356.

Changes in other repositories:
- https://github.com/ckeditor/ckeditor5-alignment/tree/t/ckeditor5-ui/344
- https://github.com/ckeditor/ckeditor5-heading/tree/t/ckeditor5-ui/344
- https://github.com/ckeditor/ckeditor5-theme-lark/tree/t/ckeditor5-ui/344
- https://github.com/ckeditor/ckeditor5-utils/tree/t/ckeditor5-ui/344

Merge commit message for related branches:

```
Internal: Aligned code to the changes in the dropdowns API (see ckeditor/ckeditor5-ui#356).
```

---

### Additional information

So this is initial PR with SplitButton (#344) and Dropdowns cleanup (#356) - mostly to open some discussion on how to close #356 as there are some things to consider here.

To quickly see how dropdowns are created now checkout [Framework Guide Stub](https://github.com/ckeditor/ckeditor5-ui/commit/fee7755828d527408f9d19b6c9a446419d3d61b6#diff-1cd2b60b887d6d0735a58a4331b24806) with two examples. This file is created mostly for PR but might be expanded if needed for guides.

To see how new dropdowns are used in features checkout:
- Alignment: https://github.com/ckeditor/ckeditor5-alignment/blob/t/ckeditor5-ui/344/src/alignmentui.js#L93-L137
- Font:
  - Family: https://github.com/ckeditor/ckeditor5-font/blob/t/ckeditor5-ui/344/src/fontfamily/fontfamilyui.js#L70-L93
  - Size: https://github.com/ckeditor/ckeditor5-font/blob/t/ckeditor5-ui/344/src/fontsize/fontsizeui.js#L65-L89
- Headings: https://github.com/ckeditor/ckeditor5-heading/blob/t/ckeditor5-ui/344/src/heading.js#L102-L126
- Highlight: https://github.com/ckeditor/ckeditor5-highlight/blob/82010b1dbb5e7c60fd67cc64d6b8e787b6775990/src/highlightui.js#L153-L232

Or checkout: `ckeditor5#t/ckeditor5-highlight/3` branch with proper `mgit.json` configuration (Sorry for branches names :trollface: but that issue evolved a bit).

As suggested I've tried to create minimal set of methods. What I like here is that button appearance is controlled by model.

---

Some things that might be improved
1. method naming (as always) and maybe location inside `ui` package
2. All features adds own CSS class - maybe make this configurable by model / helper?
```js
// example:
dropdownView.extendTemplate( {
	attributes: {
		class: 'ck-splitbutton-dropdown'
	}
} );
```
3. All features adds execute command & focus editing view afterward - helper?
```js
// Execute current action from dropdown's split button action button.
dropdownView.on( 'execute', () => {
	// This line is similar but need command name and param:
	editor.execute( 'command', { value: model.commandValue } );
	// This line is duplicated always:
	editor.editing.view.focus();
} );
```
4. I had method `addDefaultBehavior()` which called three methods but I've removed to meke less methods. Below code is duplicated everywhere:
```js
// Add default behavior of dropdown
closeDropdownOnBlur( dropdownView );
closeDropdownOnExecute( dropdownView );
focusDropdownContentsOnArrows( dropdownView );
```
5. Whether to pass toolbar/listview contents by model or by method param (I'm for the latter)
6. Toolbar - ListView dropdowns inconsistency in adding view property
```js
// for toolbar:
dropdownView.toolbarVIew;
// for listview:
dropdownView.listView;
// maybe just:
dropdownView.contentView; // ??
```

---
Things that require fixes:
- The SplitButton's CSS & button inside dropdown must be fixed (awaiting theme refactor as @oleq worked on that arrow thingy).
- "execute" event is not delegated properly for SplitButton (to check at least) - see guide both "execute" are on different elements.
- docs :D


jodator:
OK I've updated the PR with the following changes:
- instead of creating button and then dropdown there are now two functions that creates empty dropdown with given button
  - `createDropdown()` (maybe needs name change?)
  - `createSplitButtonDropdown()`
- there are no default behavior methods adding default behavior is done by above methods
- the `addListViewToDropdown()` and `addToolbarToDropdown()` now accepts `dropdown` and `items` or `buttons` respectively
- there is no `model` needed - all bindings are now on `dropdownView` directly (maybe should go on `dropdownView.button` 🤔 
- I've created `bindOneToMany()` binding

Most changes are visible on fremework guide: https://github.com/ckeditor/ckeditor5-ui/compare/fee7755...t/344#diff-1cd2b60b887d6d0735a58a4331b24806

Updated usage in features:
- Alignment: https://github.com/ckeditor/ckeditor5-alignment/blob/1f598198fde5be2853dfa76cb87eafe1c6081489/src/alignmentui.js#L86-L123
- Font Family: https://github.com/ckeditor/ckeditor5-font/blob/dd25c6e8396e5650ed98ffa9e8c838995845dd2d/src/fontfamily/fontfamilyui.js#L56-L83
- Font Size: https://github.com/ckeditor/ckeditor5-font/blob/dd25c6e8396e5650ed98ffa9e8c838995845dd2d/src/fontsize/fontsizeui.js#L50-L78
- Heading: https://github.com/ckeditor/ckeditor5-heading/blob/44627f1f9343da0dbbd96063b54ccb48ede2d084/src/heading.js#L74-L111
- Highlight: https://github.com/ckeditor/ckeditor5-highlight/blob/fdb76bc5e88cfcac9167471e9d633268338378be/src/highlightui.js#L176-L251

Reinmar:
Some notes from the call:

* Move `bindOneToMany()` to `ckeditor5-utils` as `foo.bind( 'x' ).toMany()`.
* Remember about default values of the dropdown view so we don't have to set them to `false` after creating a dropdown.
* "Not an array above" misses period. And we miss "the" a lot.
* https://github.com/ckeditor/ckeditor5-highlight/blob/d1e0d04706503bd04e507ac4c88fbf80d27ac35f/src/highlightui.js#L189 is unnecessary and it could be called simply `value`.

docs/framework/guides/dropdowns.md
```diff
@@ -0,0 +1,69 @@
+---
```
Reinmar: Can we merge this into the ui-library guide? Perhaps under this section: https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/framework/guides/architecture/ui-library.html#using-the-existing-components – we could have two subsections – buttons and dropdowns.

src/dropdown/button/dropdownbuttonview.js
```diff
@@ -0,0 +1,80 @@
...
+
+import ButtonView from '../../button/buttonview';
+
+import dropdownArrowIcon from '../../../theme/icons/dropdown-arrow.svg';
+import IconView from '../../icon/iconview';
+
+/**
+ * The default dropdown button view class.
+ *
+ *		const view = new SplitButtonView();
```
Reinmar: SplitButtonView?

src/dropdown/button/dropdownbuttonview.js
```diff
@@ -0,0 +1,80 @@
...
+ * @module ui/dropdown/button/dropdownbuttonview
+ */
+
+import ButtonView from '../../button/buttonview';
+
+import dropdownArrowIcon from '../../../theme/icons/dropdown-arrow.svg';
+import IconView from '../../icon/iconview';
+
+/**
+ * The default dropdown button view class.
```
Reinmar: Please add some links how you'd really create dropdowns. Cross-links are one of the most important thing in any documentation because they add context. Here, you showed how to create a button, but is this what people are interested in?

src/dropdown/button/splitbuttonview.js
```diff
@@ -0,0 +1,254 @@
...
+
+import KeystrokeHandler from '@ckeditor/ckeditor5-utils/src/keystrokehandler';
+import FocusTracker from '@ckeditor/ckeditor5-utils/src/focustracker';
+
+import dropdownArrowIcon from '../../../theme/icons/dropdown-arrow.svg';
+
+import '../../../theme/components/button/splitbutton.css';
+
+/**
+ * The split button view class.
```
Reinmar: The same thing as in the other button – misses links.

src/dropdown/button/splitbuttonview.js
```diff
@@ -0,0 +1,254 @@
...
+		/**
+		 * Used to create a {@link #tooltip}.
+		 *
+		 * @observable
+		 * @member {String} #label
+		 */
+		this.set( 'label' );
+
+		/**
+		 * (Optional) The keystroke associated with the button, i.e. <kbd>CTRL+B</kbd>,
```
Reinmar: Ctrl+B

src/dropdown/button/splitbuttonview.js
```diff
@@ -0,0 +1,254 @@
...
+ *			label: 'A button',
+ *			keystroke: 'Ctrl+B',
+ *			tooltip: true
+ *		} );
+ *
+ *		view.render();
+ *
+ *		document.body.append( view.element );
+ *
+ * @extends module:ui/view~View
```
Reinmar: Am I right that it has to be the same interface as `ButtonView`? It would be good to have this interface documented. Then, you wouldn't have to duplicate the API docs.

src/dropdown/dropdownpanelview.js
```diff
@@ -65,4 +65,22 @@ export default class DropdownPanelView extends View {
 			}
 		} );
 	}
+
+	focus() {
```
Reinmar: Docs?

Reinmar:
Do we need to have all these `label`, `withText`, etc properties on `DropdownView`? I assume that the button is accessible through `DropdownView#button` so one can set them directly there?

src/dropdown/dropdownview.js
```diff
@@ -158,7 +237,7 @@ export default class DropdownView extends View {
 		super.render();
 
 		// Toggle the the dropdown when it's button has been clicked.
```
Reinmar: "its"

src/dropdown/dropdownview.js
```diff
@@ -158,7 +237,7 @@ export default class DropdownView extends View {
...
-		this.listenTo( this.buttonView, 'execute', () => {
+		this.listenTo( this.buttonView, 'select', () => {
```
Reinmar: What does the `select` event mean? That a dropdown was opened?

src/dropdown/utils.js
```diff
@@ -7,56 +7,180 @@
...
- * Adds a behavior to a dropdownView that focuses dropdown panel view contents on keystrokes.
+ * A helper which creates an instance of {@link module:ui/dropdown/dropdownview~DropdownView} class with an instance of
+ * a button class passed as `ButtonClass` parameter`
+ *
+ * The default value of `ButtonClass` is {@link module:ui/dropdown/button/dropdownbuttonview~DropdownButtonView} class.
+ *
+ *		const dropdown = createDropdown( model );
+ *
+ *		// Configure dropdown properties:
+ *		dropdown.set( {
```
Reinmar: This will change to `dropdown.button.set()`, yes?

Reinmar: Or to two separate `set()` calls (on the dropdown and on the button)?

Reinmar:
> Do we need to have all these label, withText, etc properties on DropdownView? I assume that the button is accessible through DropdownView#button so one can set them directly there?

Although, when I look at the bindings in `createDropdown()` it may turn out that it won't be that easy. Anyway, curious to see the result, but I'm not sure yet whether this is the right direction.

src/dropdown/utils.js
jodator: In this case two - the `dropdown.set( 'isEnabled', true )` is required.

Reinmar: Can't `isEnabled` default to `true`?

src/dropdown/button/splitbuttonview.js
jodator: It turns out that it also plays nice with extending other interfaces. I've created `DropdownButtonInterface` that `@extends module:ui/button/buttoninterface~ButtonInterface` - event less code :)

src/dropdown/dropdownview.js
jodator: It means that "arrow" was pressed. I've introduce it to have another event on button along side `execute` because on `SplitButtonView` I need two events:
- one for `execute`
- second one for opening a panel.

It was previously `execute` event on a `selectView` of this button but I've renamed it to `arrowView` as @oleq already named similar part of UI on `DropdownButton`.

Probably it will be better to rename this event to `open`. WDYT?

jodator:
@Reinmar & @oleq I've udpated this PR with changes:
- no binding from dropdown (use `dropdown.buttonView` instaed to control button)
- introduced new interfaces for buttons
- removed `createSplitButtonDropdown()` method -> the `createDropdown()` method now accepts a `ButtonClass` parameter
- updated the docs ;)

src/button/buttoninterface.jsdoc
```diff
@@ -0,0 +1,140 @@
...
+ *			keystroke: 'Ctrl+B',
+ *			tooltip: true,
+ *			withText: true
+ *		} );
+ *
+ *		view.render();
+ *
+ *		document.body.append( view.element );
+ *
+ * @extends module:ui/view~View
```
Reinmar: What is this? ;) Is it an interface, is it a class or is it a plain?

src/button/buttoninterface.jsdoc
```diff
@@ -0,0 +1,140 @@
...
+ */
+
+/**
+ * @module ui/button/buttoninterface
+ */
+
+/**
+ * The button interface.
+ *
+ * @interface module:ui/button/buttoninterface~ButtonInterface
```
Reinmar: I don't like using the word `Interface` in names of interfaces but sometimes it's just super hard to find a way around this. There's already `Plugin` and `PluginInterface` because I wasn't able to name both without using the `Interface` suffix.

src/dropdown/button/dropdownbuttoninterface.jsdoc
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * The dropdown button interface.
+ *
+ * @interface module:ui/dropdown/button/dropdownbuttoninterface~DropdownButtonInterface
+ * @extends module:ui/button/buttoninterface~ButtonInterface
+ */
+
+/**
+ * Fired when the arrow view is clicked. It won't be fired when the button {@link #isEnabled} is `false`.
```
Reinmar: Isn't this for the split button?

src/button/buttoninterface.jsdoc
jodator: It's a copy-paste artifact.

src/dropdown/button/dropdownbuttoninterface.jsdoc
jodator: Yeah - it should be "fired when the button is clicked".

src/dropdown/utils.js
```diff
@@ -7,56 +7,164 @@
...
-			panelViewContents.focusLast();
-			cancel();
-		}
-	} );
+	const panelView = new DropdownPanelView( locale );
+	const dropdownView = new DropdownView( locale, buttonView, panelView );
+
+	buttonView.bind( 'isEnabled' ).to( dropdownView );
+
+	if ( buttonView instanceof DropdownButtonView ) {
```
Reinmar: This is ugly. Why is the `createDropdown()` function responsible for that? What if I'd create another button view which implements the same interface as the dropdown button but is not an instance of `DropdownButtonView`?

Reinmar: We should delegate this logic out of here.

src/dropdown/button/splitbuttonview.js
```diff
@@ -0,0 +1,257 @@
...
+		} );
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	render() {
+		super.render();
+
+		this.children.add( this.actionView );
```
Reinmar: For the future: https://github.com/ckeditor/ckeditor5-utils/issues/221

Reinmar:
OK, we're not going to use the word `Interface` by just stripping it. So there will be the `Button` interface and the `ButtonView` class. It's not great, but the actual correct naming would be `ButtonViewInterface` which is catastrophically long. E.g. we'd have `module:ui/dropdown/button/dropdownbuttonviewinterface~DropdownButtonViewInterface`. Let's go the other way around. These are only interface names so we can always change them later if we wish.

Reinmar:
OK, I closed this PR and branches in the utils, alignment, theme-lark and heading repos. I leave the rest to you, @jodator.

Also, I reported https://github.com/ckeditor/ckeditor5/issues/831 and plan to report issue about the `if ()` in the `createDropdown()` function.

PR #363, opened 2018-02-02
Rename isDomNode to isNode

ma2ciek:

Other: Renamed isDomNode to isNode.


Reinmar:
These kind of changes are internal since they don't affect the API of the this package.

PR #367, opened 2018-02-08
Fix: Button icon styles should not touch dropdown's arrow.

jodator:

Fix: Button icon styles should not touch dropdown's arrow. Closes ckeditor/ckeditor5#831.

---

### Additional information
Now:
selection_031


theme/components/button/mixins/_button.css
```diff
@@ -7,7 +7,7 @@
  * A helper to define button–specific icon styles.
  */
 @define-mixin ck-button-icon {
-	& .ck-icon {
+	& .ck-icon:not(.ck-dropdown__arrow) {
```
oleq: I'm not sure this is the best way to handle that. For one thing, it'll render this `:not(.ck-dropdown__arrow)` wherever the mixin is used. Even if it has nothing to do in dropdowns. It bloats the size of the build.

Another thing is that one may compile editor without `DropdownView` but still with buttons. Why would they expect dropdown–specific styles appear in their CSS?

I think it's either:
* up to the dropdown CSS to manage this case
* or we could narrow the mixin
   ```diff
     * A helper to define button–specific icon styles.
     */
    @define-mixin ck-button-icon {
   -	& .ck-icon {
   +	& .ck-icon:first-child {  // :first-of-type?
    		@mixin-content;
    	}
    }
   ```
   and simply assume that if an icon is to be an actual button icon, it goes first in DOM (it does).

The second solution feels the best. I hope it doesn't blow up.

jodator: The `first-*` would not work with dropdown button that has only text (ie heading)
`first-of-type`:
selection_038
`:not(.ck-dropdown__arrow)`:
selection_039

jodator: On the above SS: The First-of-type (`ck-icon`) is dropdown-arrow which shouldn't have that styles set.

oleq: But `first–child` works, doesn't it?
image

jodator: Or maybe just:
```css
&.ck-icon {
    margin: 0;
}
```
in here:
https://github.com/ckeditor/ckeditor5-ui/blob/0f26ca8fbf2d1e306fc0f85b12d26ff2bec573b2/theme/components/dropdown/dropdown.css#L8-L19

oleq: Alternatively, we could add a class `.ck-button svg.ck-icon.ck-button__icon` for styling actual (bold) icons.

All we need is a `className` observable in the `IconView`, and an assignment around here
https://github.com/ckeditor/ckeditor5-ui/blob/0f26ca8fbf2d1e306fc0f85b12d26ff2bec573b2/src/button/buttonview.js#L150-L154

jodator: I've updated the PR with above change.

theme/components/dropdown/dropdown.css
```diff
@@ -16,6 +16,10 @@
 		position: absolute;
 		top: 50%;
 		transform: translate3d( 0, -50%, 0 );
+
+		&.ck-icon {
```
oleq: But then, if any new style is added to the `.ck-button .ck-icon` we have a new issue and a new "undo" in this place. I'd rather go with https://github.com/ckeditor/ckeditor5-ui/pull/367#discussion_r167011336.

theme/components/button/mixins/_button.css
jodator: OH sorry... I've missed your comment about `first-child`.

jodator:
Yep - you're right. I've updated the PR once again.

PR #386, opened 2018-04-13
 Fix: Attached panel to the correct range in a case of multi-range selection.

### LINKED ISSUE #385: Incorrect BalloonToolbar position in a case of multi-range selection
In Firefox when selection contains widget then selection range is split. BalloonToolbar does not take multi-range selection into consideration and always gets the first range as a target. It causes incorrect toolbar position:

### Forward selection
image

oskarwrobel:

Fix: Attached panel to the correct range in a case of multi-range selection. Closes #385.

---

### Additional information

Manual test: http://localhost:8125/ckeditor5-ui/tests/manual/tickets/385/1.html.


Reinmar:
@oleq, could you review this PR?

oleq:
Yep, I'm checking this out.

oskarwrobel:
Any progress here? :)

PR #388, opened 2018-05-09
t/ckeditor5/920b: The action should be executed upon the first click on a tooltip-enabled UI in iOS

oleq:

Fix: The action should be executed upon the first click on a tooltip-enabled UI in iOS. Closes ckeditor/ckeditor5#920.


oleq:
Please test it in all browsers: Chrome, FF, Edge, Safari, Safari@iOS.

Issue #181, opened 2017-03-31
Panel sticks out of the container element while scroll


oskarwrobel:
Follow-up of: #170 

![mar-31-2017 12-37-33](https://cloud.githubusercontent.com/assets/1057056/24547209/ec6978b0-160e-11e7-98d8-dec905b55dbc.gif)

oleq:
And what is the limiter here?

oskarwrobel:
Editable element.

oleq:
https://github.com/ckeditor/ckeditor5-ui/issues/173 should resolve it. Precisely:

> Additionally getOptimalPosition() could check all the ancestors of the limiter which have overflow different than visible and intersect all their rects one by one up to window to find the real visible area which is available to position the BalloonPanel. That would, in most cases, make the limiter configuration obsolete.

oskarwrobel:
Problem is not with configuration. The question is what should we do with panel element when target element is out of the visible part of the limiter?

oskarwrobel:
Should we hide it or keep inside limiter bounds?

oleq:
That's exactly what I quoted ;-) 

`getOptimalPosition()` will essentially traverse up to the root of the document to check if the limiter isn't restricted by some `overflow: [scroll[,x-y]|auto]` and if so, consider that fact. That's it.

fredck:
> keep inside limiter bounds

+1

oleq:
For the record, this issue is actually more complex than https://github.com/ckeditor/ckeditor5-utils/issues/148 and not covered by the fix.

It is complex because:
1. As `#editable` is the `limiter`, there's no way to move the body collection containing the panel inside of it, so the panel gets cropped by the `overflow` of the (common) parent.
2. Because of 1., the only solution is analyzing the geometry of the panel/limiter and taking necessary actions.
3. It's a nasty business, though: suddenly `attachTo()` and `pin()` become responsible for the visibility of the panel, which is controlled by `#isVisible` attribute.
	1. `#isVisible` is the interface of the balloon. Features use it to control it. So usually it's bound to some external attribute like `isFocused` of `focusManager` (like the contextual toolbar) or something else.
	2. When `attachTo()` and `pin()` start touching `#isVisible` things get complicated. The balloon becomes hidden when it reaches the edge of the limiter. 

		But can we show it back again if the geometry allows? 

[...]

PR #392, opened 2018-05-18
Feature: Introduced BlockToolbar plugin.

### LINKED ISSUE #391: Introduce BlockToolbar
We already have a block toolbar in Letters and we decided to make it available for CKE5.

may-18-2018 09-31-47

oskarwrobel:

Feature: Introduced BlockToolbar plugin. Closes #391.

---
### Requires:
- https://github.com/ckeditor/ckeditor5-theme-lark/pull/175
- https://github.com/ckeditor/ckeditor5-utils/pull/241
- https://github.com/ckeditor/ckeditor5-core/pull/128

Constelation - https://github.com/ckeditor/ckeditor5/tree/t/ckeditor5-ui/391

### Preview

may-18-2018 13-28-39


Reinmar:
Those Saturday's nights!

Reinmar:
At least, make yourself that pleasure and listen to https://www.youtube.com/watch?v=9K_53SzflbE&index=3&list=PLM4tzdCeCG--rRO07jDfssWg2YrMu2Wsy :)

oskarwrobel:
Relaxing. Perfect for running docs validation :D

src/toolbar/block/blocktoolbar.js
```diff
@@ -0,0 +1,417 @@
...
+/**
+ * The block toolbar plugin.
+ *
+ * This plugin provides button attached to the block of content where the selection is currently placed.
+ * After clicking on the button, dropdown with editor features defined through
+ * {@link module:core/editor/editorconfig~EditorConfig#blockToolbar} appears.
+ *
+ * By default button is allowed to be displayed next to {@link module:paragraph/paragraph~Paragraph paragraph element},
+ * {@link module:list/list~List list items} and all items defined in {@link module:heading/heading~Heading} plugin.
+ * This behavior can be customise through decorable {@link ~BlockToolbar#checkAllowed} method.
```
pjasiun: We talked with @Reinmar today, we realized that we should check schema instead. Block toolbar should be displayed next to `block`s, but should not be next to `object`s.

src/toolbar/block/blocktoolbar.js
```diff
@@ -0,0 +1,423 @@
...
+			if ( isReadOnly ) {
+				this._disable();
+			} else {
+				this._enable();
+			}
+		} );
+
+		// Checking if button is allowed for displaying next to given element is event–driven.
+		// It is possible to override #checkAllowed method and apply custom validation.
+		this.decorate( 'checkAllowed' );
```
pjasiun: If we will be able to do it based on the schema we can remove this decorator and change it to the private method.

src/toolbar/block/blocktoolbar.js
```diff
@@ -0,0 +1,423 @@
...
+	/**
+	 * This event is fired when {@link #checkAllowed} method is executed. It makes it possible to override
+	 * default method behavior and provides a custom rules.
+	 *
+	 * @event checkAllowed
+	 */
+}
+
+// Because the engine.view.writer.getParentContainer is not exported here is a copy.
+// See: https://github.com/ckeditor/ckeditor5-engine/issues/628
```
pjasiun: Feel free to move this method to `view.position`.

src/toolbar/block/blocktoolbar.js
pjasiun: I just realized that there are cases where one may want to disable it for a block: for instance document title. So this method could stay.

src/toolbar/block/blocktoolbar.js
oskarwrobel: I'm for not excluding objects because when you add blockquote to the toolbar you should be able to quote and unquote an image. We can do the same thing as in `BalloonToolbar` and do not show toolbar when all items are disabled, so without blockquote toolbar won't be displayed next to the image.

oskarwrobel: When toolbar won't be displayed when all items inside are disabled then we can remove the code that checks `isReadOnly` state because in read-only all items should always be disabled so the toolbar won't be displayed anyway.

pjasiun:
There is no clear information in the manual test that there should be no balloon toolbar next to header 1. I had to check source code to ensure it is not a bug.

src/toolbar/block/blocktoolbar.js
```diff
@@ -0,0 +1,342 @@
...
+ *
+ * This plugin provides button attached to the block of content where the selection is currently placed.
+ * After clicking on the button, dropdown with editor features defined through
+ * {@link module:core/editor/editorconfig~EditorConfig#blockToolbar} appears.
+ *
+ * By default button is allowed to be displayed next to all elements marked in
+ * {@link module:engine/model/schema~Schema} as `$block` elements for which there is at least
+ * one available option in toolbar. E.g. Toolbar with {@link module:paragraph/paragraph~Paragraph} and
+ * {@link module:heading/heading~Heading} won't be displayed next to {@link module:image/image~Image} because
+ * {@link module:engine/model/schema~Schema} disallows to change format of {@link module:image/image~Image}.
```
pjasiun: This is not true anymore.

src/toolbar/block/blocktoolbar.js
```diff
@@ -0,0 +1,342 @@
...
+		} );
+
+		// Try to hide button when editor switch to read-only.
+		// Do not hide when panel was visible to avoid confusing situation when
+		// UI unexpectedly disappears.
+		this.listenTo( editor, 'change:isReadOnly', () => {
+			if ( !this.panelView.isVisible ) {
+				this.buttonView.isVisible = false;
+			}
+		} );
```
pjasiun: Does it mean that if the panel was visible the button won't disappears when the panel will be hidden? For me, it is fine if the panel will disappear when you switch to the read-only mode.

pjasiun:
Does `line-height: 1em` is not needed anymore?

src/toolbar/block/blocktoolbar.js
oskarwrobel: When editor switches to read-ony while you are using BlockButton (panel is opened) then panel won't disappear as long as you don't close it. When you close it button will disappear as well. This is to avoid situation when you want to click something on panel but pannel together with button disappear unexpectedly because of editor switches to read-only.

src/toolbar/block/blocktoolbar.js
oskarwrobel: It's still true. BlockToolbar will be not displayed next to element for which schema disallows to modify it using toolbar options. Still schema decides.

oskarwrobel:
> There is no clear information in the manual test that there should be no balloon toolbar next to header 1. I had to check source code to ensure it is not a bug.

Button should be displayed next to all headers. Button is not displayed next to image because there are no available options in toolbar for changing image.

oskarwrobel:
> Does line-height: 1em is not needed anymore?

No is not.
See https://github.com/ckeditor/ckeditor5-ui/pull/392/files#diff-5c933689b1a22af6c577f3f83c5a82b3R263.

src/toolbar/block/blocktoolbar.js
```diff
@@ -0,0 +1,342 @@
...
+		const editor = this.editor;
+
+		editor.editing.view.addObserver( ClickObserver );
+
+		/**
+		 * Toolbar view.
+		 *
+		 * @type {module:ui/toolbar/toolbarview~ToolbarView}
+		 */
+		this.toolbarView = new ToolbarView( editor.locale );
```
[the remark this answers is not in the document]
oskarwrobel: Fixed.

src/toolbar/block/view/blockbuttonview.js
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+import ButtonView from '../../../button/buttonview';
+import toUnit from '@ckeditor/ckeditor5-utils/src/dom/tounit';
+import '../../../../theme/components/toolbar/blocktoolbar.css';
+
+const toPx = toUnit( 'px' );
+
+/**
+ * The block button view class.
```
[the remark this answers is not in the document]
oskarwrobel: Fixed.

oleq:
The button fails to reposition as the target is being scrolled inside an `overflow: scroll` container.

kapture 2018-05-28 at 16 14 43

Should we handle this case? If so, even if we used the `BalloonPanelView` as button's parent, we're gonna fall into https://github.com/ckeditor/ckeditor5-ui/issues/181 anyway.

src/toolbar/block/view/blockbuttonview.js
```diff
@@ -0,0 +1,60 @@
+/**
+ * Copyright (c) 2016 - 2017, CKSource - Frederico Knabben. All rights reserved.
+ */
+
+/**
+ * @module ui/toolbar/block/view/blockbuttonview
```
oleq: Do we need this `view` subfolder? It's unlikely another view will be there and it makes the structure unnecessarily complex.

tests/toolbar/block/blocktoolbar.js
```diff
@@ -70,11 +70,11 @@ describe( 'BlockToolbar', () => {
 			} );
 
 			it( 'should add panelView to ui.focusTracker', () => {
-				expect( editor.ui.focusTracker.isFocused ).to.false;
+				expect( editor.ui.focusTracker.isFocused ).to.be.false;
```
oskarwrobel: :(

oskarwrobel:
Ticket about repositioning on scroll issue https://github.com/ckeditor/ckeditor5-ui/issues/398.

PR #407, opened 2018-06-12
t/ckeditor5-table/24: Implemented list component separators

oleq:

Feature: Implemented list component separators (see ckeditor/ckeditor5-table#24).

---

### Additional information

Part of https://github.com/ckeditor/ckeditor5-table/pull/46.

### DIFF
--- a/src/list/listseparatorview.js
+++ b/src/list/listseparatorview.js
@@ -1,0 +1,34 @@
+/**
+ * @license Copyright (c) 2003-2018, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+/**
+ * @module ui/list/listseparatorview
+ */
+
+import View from '../view';
+
+/**
+ * The list separator view class.
+ *
+ * @extends module:ui/view~View
+ */
+export default class ListSeparatorView extends View {
+	/**
+	 * @inheritDoc
+	 */
+	constructor( locale ) {
+		super( locale );
+
+		this.setTemplate( {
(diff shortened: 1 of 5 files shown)


src/list/listseparatorview.js
```diff
@@ -0,0 +1,34 @@
...
+ */
+export default class ListSeparatorView extends View {
+	/**
+	 * @inheritDoc
+	 */
+	constructor( locale ) {
+		super( locale );
+
+		this.setTemplate( {
+			tag: 'span',
```
jodator: I have doubts that it will be a valid HTML? The list items are `<li>` elements inside `<ul>` (right now they are not buttons). If we plan to switch list to set of to buttons/other elmeents then it's probably OK.

ps.: I've checked this in HTML validator:
>     Error: Element span not allowed as child of element ul in this context. (Suppressing further errors from this subtree.)
>
>    From line 8, column 308; to line 8, column 343
> 
>      left</li><span class="ck ck-list__separator"></span
> 
>     Contexts in which element span may be used:
>         Where phrasing content is expected.
>     Content model for element ul:
>         Zero or more li and script-supporting elements.

oleq: You're right. I copy-pasted it from the toolbar component. In fact... we can get rid of the `ListSeparatorView` and use `ListItemView` with an additional CSS class instead.

jodator: I find this clean to have `ListSeparatorView` instead of reausing `ListItemView` so it gonna be OK.

PR #409, opened 2018-06-13
Aligned the `BootstrapEditor` to changes in ckeditor5-core

pomek:

Other: Aligned the `BootstrapEditor` to changes related to base `Editor` class and interfaces (`EditorWithUI` and `ElementApi`).

---

Requires: https://github.com/ckeditor/ckeditor5-core/pull/129


docs/_snippets/examples/bootstrap-ui-inner.js
```diff
@@ -69,9 +69,13 @@ export default class BootstrapEditor extends Editor {
 		this._elementReplacer = new ElementReplacer();
 	}
 
+	get element() {
```
Reinmar: In this type of the editor UI `ui.view.element` is always `null`. I'm not sure whether it makes sense to pretend that `editor.element` returns something else. So, it'd be clearer to move it to the constructor and set it to `null`.

Reinmar: Oops. Wrong repo :D I'm deleting my above comment.

PR #414, opened 2018-06-22
t/402: Implemented the SwitchButtonView

### LINKED ISSUE #402: Implement the switch toggle.
The `ListView` should support adding a separator as an item.

Below is an example (by @oleq) of such toggle in list view:
State OFF:
selection_054
State ON:
selection_055

Required by [the Table UI](https://github.com/ckeditor/ckeditor5-table/issues/1#issuecomment-392034926).

As now I see it only as a list view item but I wonder if we might need it in other places?

### LINKED ISSUE #403: Consider using ButtonView inside ListItemView
As the framework develops, `ListItemView` becomes more and more like `ButtonView`:

* it `executes` when clicked,
* it has some keyboard support,
* [now it gets `isEnabled`](https://github.com/ckeditor/ckeditor5-ui/issues/389).

Soon it will get icons, tooltips, etc. and we'll have 2 components doing the same job.

Instead, I propose that we use `ButtonView` in `ListItemView` where the later is just a `<li>` DOM wrapper and an event/property proxy.

The button could have a special `type` for that purpose, or `LIV` would simply extend button's template and add a CSS class that changes its look.

cc @jodator @Reinmar @oskarwrobel

oleq:

Feature: Implemented the `SwitchButtonView`. Closes #402. Closes #403.

Also:
* Simplified the `ListItemView` class, which is now just a container for `ButtonView` (and others),
* Moved `ListItemView#style` and `#class` to the `ButtonView` (as `#class` and `#labelStyle`),
* Enhanced the `addListToDropdown` utility with a better configuration (`ListDropdownItemDefinition`) and the support for buttons, switch buttons, and separators,
* `.ck-button` and `.ck-list` became `flex` containers; the first one to allow complex inner structures (like the toggle) and the later to take control of complex list items (like `.ck-switchbutton`).

BREAKING CHANGE: Most of the `ListItemView` functionality is now handled by the `ButtonView`,
BREAKING CHANGE: The API of the `addListToDropdown` has changed; see `ListDropdownItemDefinition` to learn more,
BREAKING CHANGE: The `.ck-button` and `.ck-list` classes are using `flex` which may have an impact on rendering.

---

### Additional information

At first, I wanted to create separate PRs for #403 and #402. Then I found out to discern the changes because the first one provides a good ground for the latter. So it's 2in1, sorry ;-)

**This PR is a constellation** (`t/ckeditor5-ui/402` built on top of `t/ckeditor5-ui/403`):

* **https://github.com/ckeditor/ckeditor5/pull/1101 docs and CI**,
* https://github.com/ckeditor/ckeditor5-theme-lark/pull/182,
* https://github.com/ckeditor/ckeditor5-editor-decoupled/pull/15,
* https://github.com/ckeditor/ckeditor5-table/pull/75,
* https://github.com/ckeditor/ckeditor5-heading/pull/108,
* https://github.com/ckeditor/ckeditor5-font/pull/17

image


src/button/button.jsdoc
```diff
@@ -120,6 +120,20 @@
...
+ *
+ * @observable
+ * @member {String} #class
+ */
+
+/**
+ * (Optional) The value of the `style` attribute of the label.
+ *
+ * @observable
+ * @member {String} #style
```
jodator: Dunno if overthinking this but maybe  it should be called `labelStyle` :thinking:. WDYT?

src/dropdown/utils.js
```diff
@@ -223,7 +231,12 @@ function closeDropdownOnBlur( dropdownView ) {
 // @param {module:ui/dropdown/dropdownview~DropdownView} dropdownView
 function closeDropdownOnExecute( dropdownView ) {
 	// Close the dropdown when one of the list items has been executed.
-	dropdownView.on( 'execute', () => {
+	dropdownView.on( 'execute', evt => {
```
jodator: I wonder if this could be done better. Right now the `dropdownView` must know what items are there and what types to surpass the `'execute'` event.

Right now I can live with that though. I'm not able to think about better (shorter) general check. Ie passing some flag in event would require more work AFAICS in the `SwitchButtonView` itself. Also we delegate all the `execute` events in `dropdown/utls` methods.

So if you don't see an easy fix for that just leave it.

src/dropdown/utils.js
```diff
@@ -248,3 +261,14 @@ function focusDropdownContentsOnArrows( dropdownView ) {
...
 	} );
 }
+
+/**
+ * A definition of the list item used by the {@link module:ui/dropdown/utils~addListToDropdown}
+ * utility.
+ *
+ * @typedef {Object} module:ui/dropdown/utils~ListDropdownItemDefinition
+ *
+ * @property {String} type Either `'separator'` or `'button'`.
```
jodator: `'switchbutton'` is missing.

src/dropdown/utils.js
oleq: I agree. It's ugly. The switch button view should pass an info along with the `'execute'` event telling the dropdown to stay open. Then any UI component could pass the same kind of info and there's no need to check the `instanceof`.

There's no easy way to pass that info, though. Ideally, it should be around
```js
buttonView.delegate( 'execute' ).to( listItemView );
```
but there's no API to extend the event on the fly in that situation. We could create a high-priority listener on `buttonView#execute` and fill in the data but... this would be ugly too.

oleq:
@dkonopka Got it. Added some transition to the background too.

kapture 2018-06-22 at 13 55 47

src/button/button.jsdoc
oleq: Alright!

PR #420, opened 2018-06-27
Hide the BalloonToolbar on editable blur

### LINKED ISSUE #418: BlockToolbar and BalloonToolbar collision
### Bug
image

### Resolution
We decided to hide `BalloonToolbar` when the editable is blurred. This will help in this case because when the `BlockToolbar`' gets the focus then the `BalloonToolbar` will hide.

oskarwrobel:

Fix: Hide the BalloonToolbar on editable blur. Closes #418.

---

### Additional information

### DIFF
--- a/src/toolbar/balloon/balloontoolbar.js
+++ b/src/toolbar/balloon/balloontoolbar.js
@@ -52,8 +52,8 @@ export default class BalloonToolbar extends Plugin {
 		this.toolbarView = this._createToolbarView();
 
 		/**
-		 * Tracks focus of editable element and {@link #toolbarView}.
-		 * When both are not focused then toolbar should hide.
+		 * Tracks the focus of the {@link module:ui/editableui/editableuiview~EditableUIView#editableElement}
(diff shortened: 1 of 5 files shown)


oskarwrobel:
> This patch break the a11y feature: Alt+F10 should focus the toolbar and allow further navigation.

Fixed.

src/toolbar/balloon/balloontoolbar.js
```diff
@@ -75,19 +77,52 @@ export default class BalloonToolbar extends Plugin {
...
-		this._handleSelectionChange();
-		this._handleFocusChange();
-
 		// The appearance of the BalloonToolbar method is event–driven.
 		// It is possible to stop the #show event and this prevent the toolbar from showing up.
 		this.decorate( 'show' );
 	}
 
+	/**
+	 * @inheritDoc
```
oleq: I'm pretty sure it does a lot. A couple of words could help.

oskarwrobel: It only attaches the listeners.

PR #425, opened 2018-07-05
Used editableElement after uiReady

### LINKED ISSUE #424: Balloon Toolbar sample does not work
screen shot 2018-07-04 at 17 43 51

oskarwrobel:

Fix: Used editableElement after uiReady. Closes #424.

---

### Additional information

### DIFF
--- a/src/toolbar/balloon/balloontoolbar.js
+++ b/src/toolbar/balloon/balloontoolbar.js
@@ -59,8 +59,11 @@ export default class BalloonToolbar extends Plugin {
 		 * @type {module:utils:focustracker~FocusTracker}
 		 */
 		this.focusTracker = new FocusTracker();
-		this.focusTracker.add( this.toolbarView.element );
-		editor.once( 'uiReady', () => this.focusTracker.add( editor.ui.view.editableElement ) );
+		// Wait for the EditorUI#init. EditableElement is not available before.
+		editor.once( 'uiReady', () => {
+			this.focusTracker.add( editor.ui.view.editableElement );
+			this.focusTracker.add( this.toolbarView.element );
+		} );
 
 		/**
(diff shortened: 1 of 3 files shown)


src/toolbar/balloon/balloontoolbar.js
```diff
@@ -59,8 +59,8 @@ export default class BalloonToolbar extends Plugin {
 		 * @type {module:utils:focustracker~FocusTracker}
 		 */
 		this.focusTracker = new FocusTracker();
-		this.focusTracker.add( editor.ui.view.editableElement );
 		this.focusTracker.add( this.toolbarView.element );
+		editor.once( 'uiReady', () => this.focusTracker.add( editor.ui.view.editableElement ) );
```
oleq: Why not simply moving both `add()` to `init()` below?

oskarwrobel: I won't help. Plugin#init is before UI#init.

oleq: Alright, that's a good point 😜Then put both `FT.add()` calls in there because now it looks like this one is super-special for some reason.

tests/toolbar/balloon/balloontoolbar.js
```diff
@@ -138,6 +139,17 @@ describe( 'BalloonToolbar', () => {
 		}, 100 );
 	} );
 
+	// See https://github.com/ckeditor/ckeditor5-ui/issues/424.
+	// This test should be removed after fixing this https://github.com/ckeditor/ckeditor5-core/issues/137.
```
oleq: When we remove this test, the change with `editor.once( 'uiReady', () => {` will be left without any unit test. Is that right?

PR #432, opened 2018-07-12
T/430: The `BalloonToolbar` should not blink when the editor receives focus

### LINKED ISSUE #430: The balloon toolbar flashes in certain situations
blinking toolbar

blinking toolbar 2

balloon editor

oleq:

Type: The `BalloonToolbar` should not blink when the editor receives focus. Closes #430.

---

### Additional information

Requires a small correction in the image tests https://github.com/ckeditor/ckeditor5-image/tree/t/ckeditor5-ui/430.

### DIFF
--- a/src/toolbar/balloon/balloontoolbar.js
+++ b/src/toolbar/balloon/balloontoolbar.js
@@ -122,6 +122,8 @@ export default class BalloonToolbar extends Plugin {
 			this._toggleVisibilityDebounced( eventName, data );
 		};
 
+		// The appearance of the BalloonToolbar method is event–driven.
+		// It is possible to stop the #show event and this prevent the toolbar from showing up.
 		this.decorate( 'show' );
 	}
 
--- a/src/toolbar/balloon/balloontoolbar.js
+++ b/src/toolbar/balloon/balloontoolbar.js
@@ -253,8 +253,15 @@ export default class BalloonToolbar extends Plugin {
 		this.listenTo( this.editor.ui, 'update', () => {
 			// Don't reposition the toolbar when awaiting visibility toggle. It may cause unnecessary
 			// movement of the toolbar before it disappears.
-			if ( !this._isVisibilityTogglePending ) {
-				this._balloon.updatePosition( this._getBalloonPositionData() );
+			if ( this._isVisibilityTogglePending ) {
+				this.once( '_toggleVisibilityDebounced', () => {
+					// The toolbar could be hidden upon #_toggleVisibilityDebounced.
+					if ( this._isToolbarVisible ) {
+						this._updatePosition();
+					}
+				}, { priority: 'low' } );
+			} else {
+				this._updatePosition();
 			}
 		} );
 
@@ -274,6 +281,15 @@ export default class BalloonToolbar extends Plugin {
 			this.stopListening( this.editor.ui, 'update' );
 			this._balloon.remove( this.toolbarView );
 		}
+	}
+
+	/**
+	 * Updates the position of the toolbar.
+	 *
+	 * @private
+	 */
+	_updatePosition() {
+		this._balloon.updatePosition( this._getBalloonPositionData() );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


src/toolbar/balloon/balloontoolbar.js
```diff
@@ -83,7 +92,35 @@ export default class BalloonToolbar extends Plugin {
...
+		this._toggleVisibilityDebounced = debounce( ( eventName, data ) => {
+			this.fire( '_toggleVisibilityDebounced', eventName, data );
+		}, 10 );
+
+		/**
+		 * Fires {@link #event:_toggleVisibilityDebounced} event using `lodash#debounce`. Once executed,
+		 * it sets the {@link #_isVisibilityTogglePending} flag `true` to notify other pieces of code
+		 * that debouncing is in progress.
+		 */
+		this._fireToggleVisibilityDebounced = ( eventName, data ) => {
```
oskarwrobel: This could be a regular class method instead of an anonymous function assigned to the class property.

src/toolbar/balloon/balloontoolbar.js
```diff
@@ -186,7 +245,19 @@ export default class BalloonToolbar extends Plugin {
 
 		// Update the toolbar position when the editor ui should be refreshed.
 		this.listenTo( this.editor.ui, 'update', () => {
-			this._balloon.updatePosition( this._getBalloonPositionData() );
+			// Don't reposition the toolbar when awaiting visibility toggle. It may cause unnecessary
+			// movement just before it disappears as a result of the toggle.
+			if ( this._isVisibilityTogglePending ) {
+				this.once( '_toggleVisibilityDebounced', () => {
```
oskarwrobel: `show()` method could reposition when ballon is already shown.
This listener won't be necessary then.

src/toolbar/balloon/balloontoolbar.js
```diff
@@ -97,33 +134,45 @@ export default class BalloonToolbar extends Plugin {
...
-			// Fire internal `_selectionChangeDebounced` event to use it for showing
-			// the toolbar after the selection stops changing.
 			this._fireSelectionChangeDebounced();
+			this._fireToggleVisibilityDebounced( 'change:range', data );
 		} );
 
-		// Show the toolbar when the selection stops changing.
 		this.listenTo( this, '_selectionChangeDebounced', () => {
-			if ( this.editor.editing.view.document.isFocused ) {
+			this._fireToggleVisibilityDebounced( '_selectionChangeDebounced' );
```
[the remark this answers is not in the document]
pjasiun: Also, it is not perfectly clear to me why we need to debounce selection change since we already listen on `change:range` and have debouncing in toggle visibility. So what is missing in the documentation is the answer to 2 question:
 - what is the difference between '_selectionChangeDebounced' and 'change:range'?
 - why do we need debounce to be used 2 time?

src/toolbar/balloon/balloontoolbar.js
```diff
@@ -97,33 +134,45 @@ export default class BalloonToolbar extends Plugin {
...
+		} );
+
+		this.listenTo( this, '_toggleVisibilityDebounced', ( evt, eventName, data ) => {
+			// Hide the toolbar when the selection is changed by a direct change or has changed to collapsed.
+			if ( eventName == 'change:range' ) {
+				if ( data.directChange || selection.isCollapsed ) {
+					this.hide();
```
pjasiun: What if there are direct selection change and focus change withing 10ms? Since only the last event name is used, there will be no information that direct selection change happen and the panel will not be hidden?

This is a bigger problem, and the reason I am not sure about the `eventName` parameter. What, in general, should be used as `eventName` if multiple events happen? If we will use only the first or the last one we will lose some information which might be important. We could use an array, but then guess which of these events should be used might be also very tricky. This is why I think that it would be much better if the event will not say what was the reason for the change. The listener should decide what to do based on the editor state (check if the selection is collapsed and if the editor is focused).

Reinmar:
@oleq do you want to keep this PR alive? Or could we close it and just keep the branch? I presume you're not planning to get back to this anytime soon.

oleq:
I can still reproduce it in Letters so the issue is valid. 

It's unlikely though we're going to schedule some time to work on it anytime soon.

On the third hand, I'd hate to lose the code BC I saw there's a lot of rocket science involved (research, ASCII art :P) and this could be a nice starting point in the future.

PR #434, opened 2018-07-16
t/427: Docs: Improved the block toolbar guide by using the balloon editor instead of the classic editor to showcase the integration

### LINKED ISSUE #427: [Docs] Improve block toolbar guide
Classic editor with block toolbar does not make much sense, even if it is possible to run it there. Why should the user open the balloon toolbar if the icon could be in the main toolbar with the immediate access?

Second, guide misses the main point of introducing ballon toolbar. The problem balloon toolbar solves is: "How to insert a content (image, table) in the editor with balloon toolbar when the selection is collapsed?". If the only goal would be to change the block type, we would add these features to the balloon toolbar, and do not introduce 2 toolbars. We should mention inserting in the guide and add a feature which illustrates it (because of [this bug](https://github.com/ckeditor/ckeditor5-table/issues/27#issuecomment-396248812), the image might be better than the table).

oleq:

Docs: Improved the block toolbar guide by using the balloon editor instead of the classic editor to showcase the integration. Closes #427.

### DIFF
--- a/docs/_snippets/features/blocktoolbar.html
+++ b/docs/_snippets/features/blocktoolbar.html
@@ -1,5 +1,8 @@
 <div id="snippet-block-toolbar">
 	<h2>The three greatest things you learn from traveling</h2>
+	<figure class="image image-style-side">
+		<img src="../assets/img/umbrellas.jpg" alt="Three Monks walking on ancient temple." />
+	</figure>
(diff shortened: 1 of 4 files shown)


Reinmar:
Ekhm... 
image
I'll fix it.

PR #439, opened 2018-08-20
t/ckeditor5-media-embed/1: Allowed displayin an error message next to the LabeledInputVIew

oleq:

Feature: Allowed displayin an error message next to the `LabeledInputVIew` (see ckeditor/ckeditor5-media-embed#1).

----

**A piece of https://github.com/ckeditor/ckeditor5-media-embed/pull/2 constellation.**

### DIFF
--- a/src/inputtext/inputtextview.js
+++ b/src/inputtext/inputtextview.js
@@ -55,7 +55,8 @@ export default class InputTextView extends View {
 		this.set( 'isReadOnly', false );
 
 		/**
-		 * `true` when the field has some error.
+		 * Set to `true` when the field has some error. Usually controlled via
+		 * {@link module:ui/labeledinput/labeledinputview~LabeledInputView#errorText}.
 		 *
 		 * @observable
 		 * @member {Boolean} #hasError
@@ -87,7 +88,7 @@ export default class InputTextView extends View {
 				placeholder: bind.to( 'placeholder' ),
 				readonly: bind.to( 'isReadOnly' ),
 				'aria-invalid': bind.if( 'hasError', true ),
-				'aria-describedby': bind.to( 'ariaDesribedById' ),
+				'aria-describedby': bind.to( 'ariaDesribedById' )
 			},
 			on: {
 				input: bind.to( 'input' )
--- a/src/inputtext/inputtextview.js
+++ b/src/inputtext/inputtextview.js
@@ -64,7 +64,7 @@ export default class InputTextView extends View {
 
 		/**
 		 * The `id` of the element describing this field, e.g. when it has
-		 * some error, which helps screen readers read the error text.
+		 * some error, it helps screen readers read the error text.
 		 *
 		 * @observable
 		 * @member {Boolean} #ariaDesribedById
@@ -93,6 +93,13 @@ export default class InputTextView extends View {
 				input: bind.to( 'input' )
 			}
 		} );
+
+		/**
+		 * Fired when the user types in the input. Corresponds to the native
+		 * DOM `input` event.
+		 *
+		 * @event input
+		 */
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


src/labeledinput/labeledinputview.js
```diff
@@ -53,6 +53,14 @@ export default class LabeledInputView extends View {
 		 */
 		this.set( 'isReadOnly', false );
 
+		/**
+		 * The error text associated with the field.
+		 *
+		 * @observable
+		 * @member {String} #error
```
Reinmar: Mismatched name.

src/inputtext/inputtextview.js
```diff
@@ -55,6 +55,14 @@ export default class InputTextView extends View {
 		 */
 		this.set( 'isReadOnly', false );
 
+		/**
+		 * `true` when the field has some error.
+		 *
+		 * @observable
+		 * @member {Boolean} #error
```
Reinmar: Mismatched name.

src/labeledinput/labeledinputview.js
```diff
@@ -80,7 +88,22 @@ export default class LabeledInputView extends View {
 			},
 			children: [
 				this.labelView,
-				this.inputView
+				this.inputView,
+				{
+					tag: 'div',
+					attributes: {
```
Reinmar: I think we miss here some ARIA attribute. Won't this text be always read?

src/labeledinput/labeledinputview.js
Reinmar: Plus this is `null` now so `{String}` isn't precise.

src/labeledinput/labeledinputview.js
```diff
@@ -53,6 +53,14 @@ export default class LabeledInputView extends View {
 		 */
 		this.set( 'isReadOnly', false );
 
+		/**
+		 * The error text associated with the field.
```
Reinmar: This is unclear to me – it says nothing about the behaviour of this property. It actually says that the error text is bound to that field for the entire time. It sounds like you have to pass it on startup and it's there forever. Plus, the "associated" word means pretty much nothing in this context. We know it's associated, cause it's that field's property :P

We should be more humanly. This is, if I understand this property's purpose, more descriptive:

> A validation error text. When set, it's displayed to the user as a typical validation error message. Set it to `null` to hide it.

src/labeledinput/labeledinputview.js
```diff
@@ -115,6 +138,13 @@ export default class LabeledInputView extends View {
 		inputView.id = id;
 		inputView.bind( 'value' ).to( this );
 		inputView.bind( 'isReadOnly' ).to( this );
+		inputView.bind( 'hasError' ).to( this, 'errorText', value => !!value );
+
+		inputView.on( 'input', () => {
```
Reinmar: This should be documented. Example how to use this property is a must if inside there's such a magic.

Reinmar:
R- due to the docs.

src/labeledinput/labeledinputview.js
oleq: It wasn't the problem because the error text is dynamic. The problem was that it wasn't read by the screen reader when it should. I found out an easy solution and it should be fine now.

inv

Reinmar:
Is it good that the errorText is reset automatically on input? What if you'd like to display typical password validation that gets updated once you type. This means that you'll need to constantly updating `errorText` on `input` because the default listener will clear it every time you insert a character. That sounds bad.

cc @oskarwrobel

Reinmar:
> So DOM#input -> UI component clears the message -> Feature logic sets the new one, DOM#input -> ...

So exactly, the error gets cleared and you set it again and it gets cleared and you set it again. The validation error message blinks.

oleq:
Technically it does blink but I think it won't be noticeable. Unless, of course, the message is assigned synchronously.

Reinmar:
So, let's check this in a separate ticket. There should be a manual test. Plus, in fact, I started thinking about this when I was reading the docs which don't explain how to use `errorText` too well :P

oleq:
Referenced our discussion in https://github.com/ckeditor/ckeditor5-ui/issues/440. Let's keep it one place for a while.

PR #444, opened 2018-09-26
t/ckeditor5-media-embed/35:  Implemented LabeledInputView#infoText to display useful hints next to the input

oleq:

Feature: Implemented `LabeledInputView#infoText` to display useful hints next to the input (see ckeditor/ckeditor5-media-embed#35).

---

### Additional information

A piece of https://github.com/ckeditor/ckeditor5-media-embed/pull/45.


Reinmar:
Are we sure we want to extend this component? Shouldn't that additional info text be displayed by the feature itself? What if we'd want an info text next to some other type of input field? And is it really useful that this text is displayed by this component? Can it be done differently?

oleq:
I was thinking about that and I guess we have 3 choices:
* stick to the implementation in this PR,
* implement it on the `MediaFormView` level, which isn't very pretty because it means we have to wrap the input in some additional container, then render the info text there, which could be cumbersome because 
    * we'd have to take care about not displaying it when the field has an error (because it's way too much),
    * we'd have to maintain its styles separately so they correspond to the input error text (font, spacing); if we changed anything in the styles of the UI library, we'd need to get back to the media embed styles (+ whatever features implement info text) and update them, which is quite a hassle IMO
* implement a form manager; I'm not sure if we have an issue for that but the more forms we have, the more useful a utility to generate them would become. Such a util could take atomic components like inputs, buttons and put them together with 
   * validation,
   * additional info texts,
   * whatever magic advanced forms require

   but it feels like a lot of design work and coding especially that it's not like we need it so badly ATM

Reinmar:
OK, sounds reasonable.

oskarwrobel:
I'll look at this on Monday.

PR #448, opened 2018-10-08
Fix: Added possibility to add custom css class to dropdownview

### LINKED ISSUE #447: Setting a custom css class on dropdownview
Hi, I cant really find that it is possible to set a custom css class to the dropdown view today. or am I wrong?

lucasreppewelander:

Fix: Made it possible to set custom css class to dropdownview. Closes #447 .

---


lucasreppewelander:
I needed the ability to style dropdowns separately and as they do not get a unique id, and i was able to do this on the buttonview but not on the dropdownview.

PR #453, opened 2018-11-06
Added `.ck-read-only` class to `EditableUIView`

dkonopka:

Other: Added the `.ck-read-only` CSS class to the `EditableUIView` when `view#isReadOnly` is true. 

---

### Additional information

Part of: https://github.com/ckeditor/ckeditor5-theme-lark/pull/208

### DIFF
--- a/src/editableui/editableuiview.js
+++ b/src/editableui/editableuiview.js
@@ -40,7 +40,7 @@ export default class EditableUIView extends View {
 					'ck-editor__editable',
 					'ck-rounded-corners',
 					bind.to( 'isFocused', value => value ? 'ck-focused' : 'ck-blurred' ),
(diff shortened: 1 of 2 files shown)


src/editableui/editableuiview.js
```diff
@@ -39,7 +39,9 @@ export default class EditableUIView extends View {
 					'ck-content',
 					'ck-editor__editable',
 					'ck-rounded-corners',
-					bind.to( 'isFocused', value => value ? 'ck-focused' : 'ck-blurred' )
+					bind.to( 'isFocused', value => value ? 'ck-focused' : 'ck-blurred' ),
+					bind.to( 'isReadOnly', value => value ? 'ck-read-only' : '' )
```
oleq: I think we can shorten this a little bit

```suggestion
					bind.if( 'isReadOnly', 'ck-read-only' )
```

PR #463, opened 2019-01-23
T/ckeditor5/479: Moved the editable attributes management to the engine

oleq:

Internal: Moved the `EditableUIView` and  `InlineEditableUIView` attributes management to the engine (see ckeditor/ckeditor5#479). Closes ckeditor/ckeditor5#798.

BREAKING CHANGE: The `EditableUIView#isReadOnly` property has been removed. Use the engine `EditableElement#isReadOnly` instead.

BREAKING CHANGE: The second argument of `EditableUIView.constructor()` and `InlineEditableUIView.constructor()` became the instance of the editing view (previously an optional editable element reference).

---

### Additional information

Part of the [constellation](https://github.com/ckeditor/ckeditor5/tree/t/479).

### DIFF
--- a/src/editableui/editableuiview.js
+++ b/src/editableui/editableuiview.js
@@ -22,10 +22,10 @@ export default class EditableUIView extends View {
 	 * @param {HTMLElement} [editableElement] The editable element. If not specified, this view
 	 * should create it. Otherwise, the existing element should be used.
 	 */
-	constructor( locale, editableElement ) {
+	constructor( locale, editableElement, editingView ) {
 		super( locale );
 
-		const bind = this.bindTemplate;
+		this.editingView = editingView;
 
 		if ( editableElement ) {
 			this.element = this.editableElement = editableElement;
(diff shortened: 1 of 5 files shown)


src/editableui/editableuiview.js
```diff
@@ -91,6 +81,28 @@ export default class EditableUIView extends View {
...
+				writer.removeClass( this.isFocused ? 'ck-blurred' : 'ck-focused', viewRoot );
+			} );
+		};
+
+		this.on( 'change:isFocused', updateFocusClasses );
+		updateFocusClasses();
+	}
+
+	disableEditingRootListeners() {
+		super.disableEditingRootListeners();
```
pjasiun:
```js
super.disableEditingRootListeners();
```
This should go to `inlineeditableuiview.js`.

src/editableui/inline/inlineeditableuiview.js
```diff
@@ -37,16 +34,35 @@ export default class InlineEditableUIView extends EditableUIView {
...
+					writer.setAttribute( 'aria-label', t( 'Rich Text Editor, %0', [ this.name ] ), this.viewRoot );
+				} else {
+					writer.removeAttribute( 'aria-label', this.viewRoot );
+				}
+			} );
+		};
+
+		this.on( 'change:name', updateAriaLabelAttribute );
+		updateAriaLabelAttribute();
+	}
```
pjasiun: I can imagine a util like this:
```js
enableEditingRootListeners() {
	updateEditingViewOn( this, 'change:isFocused', ( writer, viewRoot ) => {
		if ( this.name ) {
			writer.setAttribute( 'aria-label', t( 'Rich Text Editor, %0', [ this.name ] ), viewRoot );
		} else {
			writer.removeAttribute( 'aria-label', this.viewRoot );
		}
	}
}
```

src/editableui/editableuiview.js
oleq: Sorry, seems like copy&paste dup.

src/editableui/inline/inlineeditableuiview.js
oleq: The biggest problem here is that attaching a listener does not set the initial (view root attribute) value. That was pretty neat in the UI approach because UI bindings handle this thing out-of-the-box.

src/editableui/inline/inlineeditableuiview.js
```diff
@@ -37,16 +34,35 @@ export default class InlineEditableUIView extends EditableUIView {
...
+
+	disableEditingRootListeners() {
```
pjasiun: It should not be needed if the engine handles it.

PR #477, opened 2019-04-04
Fix: Fixed View#render collision when moving focus from a one editable to the other in multi-root editor

oskarwrobel:

Fix: Fixed `View#render` collision when moving focus from a one editable to the other in multi-root editor. Closes https://github.com/ckeditor/ckeditor5/issues/1676.

---

### Additional information

### DIFF
--- a/src/editableui/editableuiview.js
+++ b/src/editableui/editableuiview.js
@@ -122,7 +122,7 @@ export default class EditableUIView extends View {
 		const editingView = this._editingView;
 
 		if ( editingView.isRenderingInProgress ) {
-			editingView.once( 'change:isRenderingInProgress', () => update( this ) );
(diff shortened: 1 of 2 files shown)


Reinmar:
Pity that you modified an existing test case. Unless it was bad, it'd make sense to have two here.

oskarwrobel:
I created two tests first. But the second one included the first one so there was no point to keep it separately.

Reinmar:
I don't buy it 😜 There could be two tests as long as they do something a bit different, and focus on bit different scenarios, even if today they boil down to the same thing, in the future they may bo covering different things.

oskarwrobel:
It's not a different scenario. It's the same scenario but the test was not good enough to find a bug. Now it is.

PR #482, opened 2019-04-11
Corrected Template#apply() docs code listing

mlewand:

Docs: Corrected `Template#apply()` docs code listing. Closes ckeditor/ckeditor5#1648.

---

### Additional information

I added more context variable to make it easy to copy'n'paste the code snippet and continue to experiment with it and removed `on` property.

Also renamed the `attrs` property into `attributes` and renamed `attrs` in `Template.bind()` which is outside of the scope of this issue, but it's not worth to make a dedicated issue for this one.

### DIFF
--- a/src/template.js
+++ b/src/template.js
@@ -158,22 +158,18 @@ export default class Template {
 	 * {@link module:ui/template~Template#revert} method.
 	 *
 	 *		const element = document.createElement( 'div' );
+	 *		const observable = new Model( { divClass: 'my-div' } );
+	 *		const emitter = Object.create( EmitterMixin );
 	 *		const bind = Template.bind( observable, emitter );
 	 *
 	 *		new Template( {
-	 *			attrs: {
+	 *			attributes: {
 	 *				id: 'first-div',
 	 *				class: bind.to( 'divClass' )
-	 *			},
-	 *			on: {
-	 *				click: bind( 'elementClicked' ) // Will be fired by the observable.
 	 *			}
-	 *			children: [
-	 *				'Div text.'
-	 *			]
 	 *		} ).apply( element );
 	 *
-	 *		element.outerHTML == "<div id="first-div" class="my-div">Div text.</div>"
+	 *		console.log( element.outerHTML ); // Logs: '<div id="first-div" class="my-div"></div>'
 	 *
 	 * @see module:ui/template~Template#render
 	 * @see module:ui/template~Template#revert
--- a/src/template.js
+++ b/src/template.js
@@ -262,7 +262,7 @@ export default class Template {
 	 *		const bind = Template.bind( observable, emitter );
 	 *
 	 *		new Template( {
-	 *			attrs: {
+	 *			attributes: {
 	 *				// Binds the element "class" attribute to observable#classAttribute.
 	 *				class: bind.to( 'classAttribute' )
 	 *			}
--- a/src/template.js
+++ b/src/template.js
@@ -1207,14 +1207,14 @@ function normalize( def ) {
 //			}
 //		}
 //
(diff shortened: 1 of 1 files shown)


src/template.js
```diff
@@ -158,22 +158,18 @@ export default class Template {
...
 	 *		const bind = Template.bind( observable, emitter );
 	 *
 	 *		new Template( {
-	 *			attrs: {
+	 *			attributes: {
 	 *				id: 'first-div',
 	 *				class: bind.to( 'divClass' )
-	 *			},
-	 *			on: {
-	 *				click: bind( 'elementClicked' ) // Will be fired by the observable.
```
oleq: Why was this line removed? [AFAIR you can add listeners using `Template.apply()`](https://github.com/ckeditor/ckeditor5-ui/blob/master/tests/template.js#L999-L1099).

mlewand:
> Also, I noticed that `attrs:` is used in other places (docs) in `template.js`. This should be addressed in this PR.

I guess you meant [this private API](https://github.com/ckeditor/ckeditor5-ui/blob/0b13eb1/src/template.js#L1211)? In that case it is updated in a57cc3e.

There are quite a few issues in our codebase (quick search shown most of them in engine module), like:

* [`engine/view/attributeelement#AttributeElement`](https://github.com/ckeditor/ckeditor5-engine/blob/9a40550a7425670264974edc9252f3dcf888ce2f/src/view/attributeelement.js#L37) (public API)
* [`engine/model/text#Text`](https://github.com/ckeditor/ckeditor5-engine/blob/9a40550a7425670264974edc9252f3dcf888ce2f/src/model/text.js#L36) (public API)
* [`engine/view/editableelement#EditableElement`](https://github.com/ckeditor/ckeditor5-engine/blob/9a40550a7425670264974edc9252f3dcf888ce2f/src/view/editableelement.js#29) (public API)
* [`DocumentSelection#_setAttributesTo`](https://github.com/ckeditor/ckeditor5-engine/blob/9a40550a7425670264974edc9252f3dcf888ce2f/src/model/documentselection.js#981) (encapsulated API)

But TBH it's ouside of the scope of this issue, as it was limited to broken code listing.

src/template.js
Reinmar: I think that removing `on` was unnecessary. It was super clear before that you could use it. Now it isn't. So it's a change in the wrong direction. Being DRY isn't the primary goal in the docs (nor in the code). The primary goal is always clarity.

src/template.js
```diff
@@ -158,22 +158,18 @@ export default class Template {
...
 	 *			}
-	 *			children: [
-	 *				'Div text.'
-	 *			]
 	 *		} ).apply( element );
 	 *
-	 *		element.outerHTML == "<div id="first-div" class="my-div">Div text.</div>"
+	 *		console.log( element.outerHTML ); // Logs: '<div id="first-div" class="my-div"></div>'
```
Reinmar: We use a simpler "->" in most places (instead of "Logs:")

Reinmar:
Just to be sure – it's impossible to pass `children` to `apply()`? It won't have any effect?

Reinmar:
Actually, I can see  you can apply such template:

```js

			it( 'overrides textContent of an existing Text Node in a HTMLElement', () => {
				el.textContent = 'bar';

				new Template( {
					tag: 'div',
					children: [ 'foo' ]
				} ).apply( el );

				expect( normalizeHtml( el.outerHTML ) ).to.equal( '<div>foo</div>' );
			} );
```

So, I'm again against removing this from the example because it becomes less explanatory.

oleq:
You can only override `textContent` using apply. It won't create or remove children.

PR #489, opened 2019-04-19
Add withArrow option to ContextualBalloon#add() method.

### LINKED ISSUE #487: It should be possible to open ContextualBalloon without arrow.
The `ContextualBalloon` API allows to pass custom positions but it lacks the ability to show/hide arrow for it. At the moment all features uses arrow but `MentionUI` does not. While working on https://github.com/ckeditor/ckeditor5-mention/issues/27 I've realised that `MentionUI` should use `ContextualBalloon` as well to work well with other features.

Proposal:

1. Add `withArrow` property to the balloon positions:

```js
BalloonPanelView.defaultPositions = {
	// ------- North

	northArrowSouth: ( targetRect, balloonRect ) => ( {
		top: getNorthTop( targetRect, balloonRect ),
		left: targetRect.left + targetRect.width / 2 - balloonRect.width / 2,
		name: 'arrow_s',
		withArrow: true
	} ),

	// ...
};
```

So the default value would be `false` - but it will be a breaking change AFAICS. The backward compatible change would be to default it to `true` but the code might get a bit less elegant.

2. Add switch to `ContextualBalloon.add()` options:

```js
this._balloon.add( {
	view: this._mentionsView,
	position: { /* position data  */ } ),
	withArrow: false
} );
```

/cc @oleq WDYT

jodator:

Other: The `ContextualBalloon` plugin's `add` method now accepts the `withArrow` option. Closes #487.

---

### Additional information

I've assumed that:
* The `withArrow` name should be the same as `BalloonPanelView` property.
* The default value is `true` for backwards compatibilty.

The PR is required for: https://github.com/ckeditor/ckeditor5-mention/pull/60 (to show/hide the arrow).

### DIFF
--- a/src/panel/balloon/contextualballoon.js
+++ b/src/panel/balloon/contextualballoon.js
@@ -115,7 +115,7 @@ export default class ContextualBalloon extends Plugin {
 	 * @param {module:ui/view~View} [data.view] Content of the balloon.
 	 * @param {module:utils/dom/position~Options} [data.position] Positioning options.
 	 * @param {String} [data.balloonClassName] Additional css class for {@link #view} added when given view is visible.
-	 * @param {Boolean} [data.withArrow=true] Whether the {#_balloon} view should be rendered with arrow.
+	 * @param {Boolean} [data.withArrow=true] Whether the balloon should be rendered with an arrow.
 	 */
 	add( data ) {
 		if ( this.hasView( data.view ) ) {
--- a/src/panel/balloon/contextualballoon.js
+++ b/src/panel/balloon/contextualballoon.js
@@ -204,7 +204,7 @@ export default class ContextualBalloon extends Plugin {
 	 * @param {Object} data Configuration.
 	 * @param {module:ui/view~View} [data.view] View to show in the balloon.
 	 * @param {String} [data.balloonClassName=''] Additional class name which will added to the {#_balloon} view.
-	 * @param {Boolean} [data.withArrow=true] Whether the {#_balloon} view should be rendered with arrow.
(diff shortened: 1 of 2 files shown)


src/panel/balloon/contextualballoon.js
```diff
@@ -115,6 +115,7 @@ export default class ContextualBalloon extends Plugin {
 	 * @param {module:ui/view~View} [data.view] Content of the balloon.
 	 * @param {module:utils/dom/position~Options} [data.position] Positioning options.
 	 * @param {String} [data.balloonClassName] Additional css class for {@link #view} added when given view is visible.
+	 * @param {Boolean} [data.withArrow=true] Whether the {#_balloon} view should be rendered with arrow.
```
oleq:
```suggestion
	 * @param {Boolean} [data.withArrow=true] Whether the balloon should be rendered with an arrow.
```

src/panel/balloon/contextualballoon.js
```diff
@@ -203,9 +204,11 @@ export default class ContextualBalloon extends Plugin {
 	 * @param {Object} data Configuration.
 	 * @param {module:ui/view~View} [data.view] View to show in the balloon.
 	 * @param {String} [data.balloonClassName=''] Additional class name which will added to the {#_balloon} view.
+	 * @param {Boolean} [data.withArrow=true] Whether the {#_balloon} view should be rendered with arrow.
```
oleq:
```suggestion
	 * @param {Boolean} [data.withArrow=true] Whether the {@link #_balloon} view should be rendered with an arrow.
```

jodator:
I am so bad with articles...

PR #497, opened 2019-05-15
Extend ColorGridView with options required in Document Colors 

msamsel:

Other: Extend ColorGridView with options required in Document Colors.

---

### Additional information

* PR required by ckeditor/ckeditor5-font#39

### DIFF
--- a/src/colorgrid/colorgridview.js
+++ b/src/colorgrid/colorgridview.js
@@ -35,7 +35,6 @@ export default class ColorGridView extends View {
 		super( locale );
 
 		const colorDefinitions = options && options.colorDefinitions || [];
-		const viewStyleAttribute = {};
 
 		/**
 		 * Number of columns in the color grid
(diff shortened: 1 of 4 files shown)


mlewand:
While reviewing at Friday I saw that it would be actually wrong to add this to ColorGridView. ColorGridView class should not be aware of such feature as a label.

Instead it's a parent responsibility to add a label before the ColorGridView in it's `children` list.

I still have to look though this PR so I'll provide a full review feedback later today.

msamsel:
> Good thing is that we already have Label class in our codebase 🎉

Yes we have it. I analyzed this feature with this case. Unfortunately it's an HTML Label:
https://github.com/ckeditor/ckeditor5-ui/blob/e2f869ee081a7cec0b6a744dd000e818c32de344/src/label/labelview.js#L45
Which AFAIK is not suitable for `<div>` element, that's why I wasn't use it in this case.

mlewand:
@msamsel It's OK to use a `label` element for labels. Ideally we'd like to have it with a proper `[for]` attribute, however our grid component doesn't have a complete accessibility markup coverage.

Once that's done the `[for]` attribute should point to a referenced grid element.

Anyway it's better to have a `label` than non-semantic element like `span` or `div`.

PR #498, opened 2019-05-15
Introduced FakePanels for ContextualBalloon.

### LINKED ISSUE #501: Improve UX of ContextualBalloon with multiple stack
We decided to improve the visual part of ContectualBalloon when more than one stack is created in the ballon.

oskarwrobel:

Other: Improved UX of ContextualBalloon with multiple stacks. Closes #501.

---

### Additional information

Required: `ckeditor5-theme-lark#cf/2808` https://github.com/ckeditor/ckeditor5-theme-lark/pull/229

### DIFF
--- a/tests/manual/contextualballoon/contextualballoon.md
+++ b/tests/manual/contextualballoon/contextualballoon.md
@@ -8,4 +8,12 @@
 ## Multiple stacks
 
 1. Select some highlighted text - "View in separate stack." should show up.
-2. Switch stacks by clicking navigation buttons - you should switch between toolbar and custom view.
+2. Switch stacks by clicking navigation buttons. You should switch between toolbar and custom views. Toolbar view should be displayed without information about number of stacks (only left and right arrows).
+
+## Fake panels - min
+
+1. Put the selection before the highlight, start moving selection by right arrow. You should see additional layer under the balloon only when at least 2 stacks are added to the balloon.
+
+## Fake panels - max
+
+1. Select text `[select]` from the lower highlight. You should see `1 of 5` status of pagination but only 3 additional layers under the balloon should be visible.
--- a/tests/manual/contextualballoon/contextualballoon.md
+++ b/tests/manual/contextualballoon/contextualballoon.md
@@ -8,7 +8,7 @@
 ## Multiple stacks
 
 1. Select some highlighted text - "View in separate stack." should show up.
-2. Switch stacks by clicking navigation buttons. You should switch between toolbar and custom views. Toolbar view should be displayed without information about number of stacks (only left and right arrows).
+2. Switch stacks by clicking navigation buttons. You should switch between toolbar and custom views.
 
 ## Fake panels - min
 
@@ -16,4 +16,4 @@
 
 ## Fake panels - max
 
-1. Select text `[select]` from the lower highlight. You should see `1 of 5` status of pagination but only 3 additional layers under the balloon should be visible.
+1. Select text `[select]` from the lower highlight. You should see `1 of 4` status of pagination but only 2 additional layers under the balloon should be visible.
--- a/tests/manual/contextualballoon/contextualballoon.md
+++ b/tests/manual/contextualballoon/contextualballoon.md
@@ -16,4 +16,4 @@
 
 ## Fake panels - max
 
-1. Select text `[select]` from the lower highlight. You should see `1 of 4` status of pagination but only 2 additional layers under the balloon should be visible.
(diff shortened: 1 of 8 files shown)


oskarwrobel:
@dkonopka JS part is done. You can make it beautiful now :)

dkonopka:
I've added some default styles - polishing in progress.

One thing to fix - we should display one more less fake panels.

Second thought: shouldn't it be named as `fakePanel` ? 

### We need to remove here one fake panel
Screenshot 2019-05-16 at 09 56 44

Screenshot 2019-05-16 at 09 55 49

pjasiun:
It might be because of the shadow, but the first the space between the real panel and the first fake panel looks smaller than spaces between other fake panels:

Screenshot 2019-05-16 at 11 30 34

Maybe we should add shadow to all panels or move it to the last one?

Also, I believe we need some animation to make the impression that panels switches. Maybe they should move behind the main panel and back to their positions when switching? Or... something. Anything ;)


fake-panels-needs-animation


Also, maybe we should try to add some transparency to these fake panels?

pjasiun:
Also, I think you should create a dedicated ticket in https://github.com/ckeditor/ckeditor5-ui/ repository for this issue.

dkonopka:
### Opacity for panels

I was trying to make it bigger/smaller, but it just looks bad:
Screenshot 2019-05-16 at 16 08 46

### Weird offset of the first panel

It was caused because additional annotation `box-shadow`, I've removed it and it looks a bit better. 

The second small reason of the wrong position of fake panels is that precision of `top`/`left` property (I mean values after comma 931.378px, there was some issue about rounding it to 931px - because Chrome and Firefox are rendering it a bit differently).

Screenshot 2019-05-16 at 16 01 28

### Anyway, it looks good right now:


fake-panels

pjasiun:
Maybe we could even limit it to 3 panels (1 real panel and 2 fake panels). Now, when 4 panels are displayed (1 real and 3 fake) it looks like too many.

Reinmar:
I'm a little late to the party, but is it actually necessary that we display those other balloons at all? After all, there's a switcher at the top so it's hard to miss that. Plus, it's not actually super correct how we display them (e.g. that all of them change their height when you go to the next one and that their placement is a bit virtual too). 

Don't get me wrong – it's a really nice touch and could be made really useful (e.g. clicking/hovering the ones below could somehow show a preview of their content). I'm rather worried whether this isn't an overkill for now.

oskarwrobel:
TBH I'm not a fan of these panels in the current shape too. Besides, it doesn't look good for small balloons like a toolbar.

pjasiun:
We agreed that we will:
 - limit the number of fake panels to 2 (1 real panel + 2 fake panel),
 - make the space between panel smaller, more subtele,
 - round the position of the main panel to full pixels to fix the issue with misaligned panels,
 - do not remove "x of x" label for small balloon (it will make them bigger and maybe panels will look there better too),
 - add an animation of appearing these panels.

pjasiun:
⏫ cc @dkonopka

theme/components/panel/fakepanel.css
```diff
@@ -0,0 +1,27 @@
...
+	z-index: 3;
+}
+
+.ck .ck-fake-panel div:nth-child( 2 ) {
+	z-index: 2;
+}
+
+.ck .ck-fake-panel div:nth-child( 3 ) {
+	z-index: 1;
+}
```
pjasiun: Do we still need 3 etries if we limit the number of fake panels to 2?

oskarwrobel: Good catch.

tests/manual/contextualballoon/contextualballoon.md
```diff
@@ -8,4 +8,12 @@
...
-2. Switch stacks by clicking navigation buttons - you should switch between toolbar and custom view.
+2. Switch stacks by clicking navigation buttons. You should switch between toolbar and custom views.
+
+## Fake panels - min
+
+1. Put the selection before the highlight, start moving selection by right arrow. You should see additional layer under the balloon only when at least 2 stacks are added to the balloon.
+
+## Fake panels - max
+
+1. Select text `[select]` from the lower highlight. You should see `1 of 4` status of pagination but only 2 additional layers under the balloon should be visible.
```
pjasiun: > You should see `1 of 4` status of pagination

I see `1 of 3` :/

oskarwrobel: image

PR #500, opened 2019-05-20
Add label for editor's toolbar

msamsel:

Fix: Add label for editor's toolbar. Closes ckeditor/ckeditor5#1404.

BREAKING CHANGE: `ToolbarView` class will now have required `locale` argument, to translate the aria-label attribute of a toolbar element.

## Additional information
Related PR: ckeditor/ckeditor5-widget#91


src/toolbar/toolbarview.js
```diff
@@ -131,6 +154,8 @@ export default class ToolbarView extends View {
 			this.focusTracker.remove( item.element );
 		} );
 
+		this.element.appendChild( this.labelView.element );
```
Reinmar: I don't think this is safe. In the same element we keep the collection of the items. In fact, the first thing that drew my attention was – why are you rendering the label separately and not in the base template.

Finally, do we need a separate element for that label? Can't we inline that label in the aria attribute?

msamsel:
I've updated this PR and made a static label for a toolbar. I also provide an option to override the aria-label, what is used by other view components which use `ToolbarView` class.

src/toolbar/toolbarview.js
```diff
@@ -27,12 +27,23 @@ import { attachLinkToDocumentation } from '@ckeditor/ckeditor5-utils/src/ckedito
  */
 export default class ToolbarView extends View {
 	/**
-	 * @inheritDoc
+	 * Creates an instance of the {@link module:ui/view~View} class.
+	 *
+	 * Also see {@link #render}.
+	 *
+	 * @param {module:utils/locale~Locale} locale The localization services instance.
+	 * @param {Object} [config]
```
oleq: No config is needed. Properties are fine.

src/dropdown/utils.js
```diff
@@ -123,13 +123,18 @@ export function createDropdown( locale, ButtonClass = DropdownButtonView ) {
...
  * See {@link module:ui/dropdown/utils~createDropdown} and {@link module:ui/toolbar/toolbarview~ToolbarView}.
  *
  * @param {module:ui/dropdown/dropdownview~DropdownView} dropdownView A dropdown instance to which `ToolbarView` will be added.
  * @param {Iterable.<module:ui/button/buttonview~ButtonView>} buttons
  */
 export function addToolbarToDropdown( dropdownView, buttons ) {
-	const toolbarView = dropdownView.toolbarView = new ToolbarView();
+	const locale = dropdownView.locale;
+	const t = locale.t;
+	const toolbarView = dropdownView.toolbarView = new ToolbarView( locale, { ariaLabel: t( 'Dropdown\'s toolbar' ) } );
```
oleq: Where are the tests?

PR #517, opened 2019-07-05
Improve the functionality of `aria-pressed` attribute in button elements

msamsel:

Fix: Improve the functionality of `aria-pressed` attribute in button elements. Closes ckeditor/ckeditor5#1403.

---

### Additional information

sub-pr:
* https://github.com/ckeditor/ckeditor5-alignment/pull/44
* https://github.com/ckeditor/ckeditor5-basic-styles/pull/95
* https://github.com/ckeditor/ckeditor5-block-quote/pull/40
* https://github.com/ckeditor/ckeditor5-heading/pull/128
* https://github.com/ckeditor/ckeditor5-highlight/pull/43
* https://github.com/ckeditor/ckeditor5-image/pull/303
* https://github.com/ckeditor/ckeditor5-link/pull/246
* https://github.com/ckeditor/ckeditor5-list/pull/135
* https://github.com/ckeditor/ckeditor5-paragraph/pull/46

### DIFF
--- a/src/button/button.jsdoc
+++ b/src/button/button.jsdoc
@@ -96,7 +96,7 @@
  */
 
 /**
- * Controls whether the button view is recognized as toggle button for assistive technologies.
+ * Controls whether the button view is a toggle button (two–state) for assistive technologies.
  *
  * @observable
  * @default false
--- a/src/button/button.jsdoc
+++ b/src/button/button.jsdoc
@@ -96,6 +96,14 @@
  */
 
 /**
+ * Controls whether the button view is recognized as toggle button for assistive technologies.
+ *
+ * @observable
+ * @default false
+ * @member {Boolean} #isToggleable
+ */
+
+/**
  * (Optional) Controls whether the label of the button is hidden (e.g. an icon–only button).
  *
  * @observable
--- a/src/button/buttonview.js
+++ b/src/button/buttonview.js
@@ -37,13 +37,7 @@ import '../../theme/components/button/button.css';
  */
 export default class ButtonView extends View {
 	/**
-	 * Creates an instance of the {@link module:ui/view~View} class.
-	 *
-	 * Also see {@link #render}.
-	 *
-	 * @param {module:utils/locale~Locale} [locale] The localization services instance.
-	 * @param {Object} [config] additional configuration
-	 * @param {Boolean} [config.toggle=false] set button as a toggle
+	 * @inheritDoc
 	 */
 	constructor( locale ) {
 		super( locale );
--- a/src/button/buttonview.js
+++ b/src/button/buttonview.js
@@ -37,9 +37,15 @@ import '../../theme/components/button/button.css';
  */
 export default class ButtonView extends View {
 	/**
-	 * @inheritDoc
+	 * Creates an instance of the {@link module:ui/view~View} class.
+	 *
+	 * Also see {@link #render}.
+	 *
+	 * @param {module:utils/locale~Locale} [locale] The localization services instance.
+	 * @param {Object} [config] additional configuration
+	 * @param {Boolean} [config.toggle=false] set button as a toggle
 	 */
-	constructor( locale ) {
+	constructor( locale, { toggle = false } = {} ) {
 		super( locale );
 
 		const bind = this.bindTemplate;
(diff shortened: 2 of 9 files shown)


msamsel:
> Your proposal adds this attribute to every button and not all buttons are toggle buttons.

I've modified existing behaviour, which has added `aria-pressed="true"` to every button. So as I understand this attribute should be added **only** for toggleable buttons.

I see 2 ways how to achieve that:
* provide configuration object passed to the constructor which define button as a toggle
* create child class `ToggleButton` which will extend `Button`

Both approaches require multiple changes in many repositories, that's why I'd like to have sure that the chosen solution will be better.

As the first one seems to be a little bit better for me, that's why I've prepared some sample solution:
https://github.com/ckeditor/ckeditor5-ui/pull/517/commits/5edce86e495aecf6899424116088bf1a18b930fa

@oleq can you give me feedback related to this case? Which approach seems to be better for this case or maybe there is some other way?

// cc @mlewand

mlewand:
BTW we already have a [`SwitchButtonView`](https://ckeditor.com/docs/ckeditor5/latest/api/module_ui_button_switchbuttonview-SwitchButtonView.html) class. Which acts like toggle button but looks differently. Maybe we could reuse `SwitchButtonView` and allow it to take different visual forms?

The only issue is that IMHO _switch button_ implies a slightly different semantics than _toggle button_. _Toggle button_ for me is more of a boolean state, always on/off, while _switch button_ sounds like a choice between two values, e.g. "wide/narrow", "black/white", "black/pink" etc.

msamsel:
> Maybe we could reuse SwitchButtonView and allow it to take different visual forms?

This sounds like overkill for me. To force the switch button, which inheritance from the button, to have a different visual form which will look like a regullar button but with defined `aria-pressed` attribute.

I think better could be to have the Button class with configuration option `toggle` and the SwitchButton class which inheritance from Button and always call `super.constructor()` with toggle configuration option set to `true`.

oleq:
KISS. Cake recipe:

1. Let's create an observable property `ButtonView#isToggleable`. Document what impact it has on the button.
2. Let's bind 
    ```js
    'aria-pressed': bind.to( 'isOn', value => this.isToggleable ? String( value ) : false )
    ```
3. Let's set `isToggleable` in places like `BoldUI`, `LinkUI`, etc. (but not `UndoUI`)
4. Let's set `this.isToggleable = true` in `SwitchButtonView#constructor()`

We don't need new constructor arguments. We incline towards observables in the project because one can change them during the life cycle of the component.

We don't need additional clas chains like `ButtonView` -> `ToggleButtonView` -> `SwitchButtonView`. There's no need to create a separate class (`ToggleButtonView`) just to add an attribute to the template. The class would be almost empty. This is not the right way to OOP.

msamsel:
Ad 2.
If I understand correctly previous comment: https://github.com/ckeditor/ckeditor5-ui/pull/517#pullrequestreview-262400785 we should not add `aria-pressed` attribute for non-toggleable buttons. @mlewand please correct me if I wrong.

oleq:
When the `bind.to()` callback returns `false` (Boolean, not String), the attribute will not be present in DOM.

mlewand:
> we should not add `aria-pressed` attribute for non-toggleable buttons

That's correct, we must not add `aria-pressed` attr for non-toggle buttons.

```js
'aria-pressed': bind.to( 'isOn', value => this.isToggleable ? String( value ) : false )
```

I'm not proficient yet with this API but my understanding is that if `bind.to` gets false it does not add an attribute, which is fine.

> We don't need additional clas chains like `ButtonView` -> `ToggleButtonView` -> `SwitchButtonView`.

My suggestion was not necessarily add another class here. Nonetheless additional property is ok.

oleq:
Just to make things clear, buttons that are not toggleable (like undo, redo, etc.) should not have `aria-pressed` at all, right @Comandeer?

> If the attribute is omitted or set to its default value of aria-pressed="undefined", the element does not support being pressed.

via https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role.

mlewand:
> should not have `aria-pressed` at all

Yes they must not have `aria-pressed` attr. That's the cause of R- in https://github.com/ckeditor/ckeditor5-ui/pull/517#pullrequestreview-262400785

The `"undefined"` value is the default value of this attribute, so there's no benefit in setting it (https://www.w3.org/TR/wai-aria-1.1/#aria-readonly).

msamsel:
I add observable `isToggleable` property to `ButtonView` and create fixes in other repositories to use it.

PR #519, opened 2019-07-10
Removed usage of logger.

ma2ciek:

Internal: Removed usage of the CKEditor 5 logger. Part of ckeditor/ckeditor5#383.

---

### Additional information

### DIFF
--- a/src/template.js
+++ b/src/template.js
@@ -7,9 +7,9 @@
  * @module ui/template
  */
 
-/* global document, console */
+/* global document */
 
-import CKEditorError, { attachLinkToDocumentation } from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
+import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
 import mix from '@ckeditor/ckeditor5-utils/src/mix';
 import EmitterMixin from '@ckeditor/ckeditor5-utils/src/emittermixin';
 import View from './view';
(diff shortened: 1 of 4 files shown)


src/template.js
```diff
@@ -381,7 +380,9 @@ export default class Template {
 			 *
 			 * @error template-extend-render
 			 */
-			log.warn( 'template-extend-render: Attempting to extend a template which has already been rendered.' );
+			console.warn( attachLinkToDocumentation(
```
pjasiun: This should be `// @if CK_DEBUG //`.

src/toolbar/toolbarview.js
```diff
@@ -180,10 +182,9 @@ export default class ToolbarView extends View {
 				 * @error toolbarview-item-unavailable
 				 * @param {String} name The name of the component.
 				 */
-				log.warn(
+				console.warn( attachLinkToDocumentation(
```
pjasiun: Technically speaking it should be an exception, because it is caused by incorrect configuration. However, for a better DX we could keep it as `console.warn`. @Reinmar: WDYT?

src/template.js
pjasiun: But to honest, I am not sure about it. If we have no ideas about cases where such extending is correct, I believe we should change it to `CKEditorError`.

pjasiun: I confirmed it with @oleq and @oskarwrobel and we agreed that it should be an exception.

src/toolbar/toolbarview.js
ma2ciek: I think that the DX is the most important thing here, so if the `console.warn` improves DX we should keep it.

pjasiun: Yea, on a second thought, I agree that we should keep `console.warn` here and in all similar cases.

PR #522, opened 2019-08-08
ColorGrid should also determine isOn value when new ColorTile is added to collections.

msamsel:

Fix: ColorGrid should also determine `isOn` value when new ColorTile is added to collections.

---

### Additional information

* issue found while testing ckeditor/ckeditor5-font#51

### DIFF
--- a/tests/colorgrid/colorgridview.js
+++ b/tests/colorgrid/colorgridview.js
@@ -106,6 +106,23 @@ describe( 'ColorGridView', () => {
 			expect( view.items.get( 2 ).isOn ).to.be.false;
 		} );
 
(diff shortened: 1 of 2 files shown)


tests/colorgrid/colorgridview.js
```diff
@@ -106,6 +106,23 @@ describe( 'ColorGridView', () => {
 			expect( view.items.get( 2 ).isOn ).to.be.false;
 		} );
 
+		it( 'should determins isOn value when colorTile is added', () => {
```
oleq:
```suggestion
		it( 'should determine #isOn value when a ColorTileView is added', () => {
```

PR #530, opened 2019-11-21
Introduced select UI component

mlewand:

Feature: Introduced the `SelectView` component. Closes ckeditor/ckeditor5#5770.

---

### Additional information

I didn't implement `isReadOnly` property as `readonly` is not supported on `select` elements in HTML so that would require workarounds.

Also I was wondering whether [changing value to a value that is not contained by the select](https://github.com/ckeditor/ckeditor5-ui/blob/7ae80a0/tests/select/selectview.js#L73-L77) should throw an error or silently fail, doing no harm. I went with the latter, though it would be nice to log some information, but I believe we have no common mechanisms for that.

There's a subpr with manual test to this component: ckeditor/ckeditor5-theme-lark#253


oleq:
We went with a `DropdownView` in https://github.com/ckeditor/ckeditor5/issues/1110 in this iteration to speed things up and use a component that is known to us. Whether we will use this component in the future, it's still unclear to the team and it needs to be discussed. For now, I'm unplugging it from the iteration.

mlewand:
Closing the PR, see https://github.com/ckeditor/ckeditor5/issues/5770#issuecomment-560396287

Issue #317, opened 2017-10-02
[Edge/FF] Exception after dragging the selection through the image


Mgsy:
## Steps to reproduce
1. Go to https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/examples/builds/balloon-editor.html.
2. Start the selection from the end of the first header.
3. Drag it down through the image.

## Current result
- The toolbar appears in the left corner of the browser.
- There is the exception in the console.

## Error
```
SCRIPT438: Object doesn't support property or method 'getBoundingClientRect'
rect.js (365,1)
```

## GIF
![bug_cke5](https://user-images.githubusercontent.com/20988892/31072632-f12d4804-a768-11e7-810a-0f2dcc4aeaf0.gif)

## Other information
OS: Windows 10
Browser: Edge 16, Firefox
Release: 0.11.0

ma2ciek:
In FF the selection starts at the end of the text that's inside the header, so in the https://github.com/ckeditor/ckeditor5-utils/blob/ac06329334e9f981e10df18237a986b56a43112e/src/dom/rect.js#L365, the `startContainer` variable is the Text element.

ma2ciek:
( `range.getClientRects()` returns an empty array. )

ma2ciek:
Maybe it can be solved by using the parent element instead of the original element when the `startContainer` is a text node.

Reinmar:
But why isn't this throwing in all browsers?

ma2ciek:
It's interesting that the error doesn't occur in the manual tests: http://localhost:8125/ckeditor5-image/tests/manual/caption.html.

ma2ciek:
When the selection starts inside the header, its start moves to the caption during dragging down. When dragging up all works fine.

ma2ciek:
Actually, this is caused by the BalloonEditor which calculates the rect for its position. It can be reproduced here too http://localhost:8125/ckeditor5-editor-balloon/tests/manual/ballooneditor.html

ma2ciek:
It's connected to https://github.com/ckeditor/ckeditor5-engine/issues/1156. There's a moment before the crash when selection starts at the end of the text and ends in the main editable element just after the text node. So there's no native rectangle available from `getClientRects` method and `startContainer` of the range is a text node, which doesn't have a `getBoundingClientRect` method.

oleq:
[...]

PR #532, opened 2019-11-29
`config` may not exist

Eusen:
In my repo. [ckeditor5-build-full](https://github.com/Eusen/ckeditor5-build-full)
I've mixed five kinds of editors together to distinguish which one should be created by the type:
 `create(type, el, config)`.
Because in a real project, it is very common to mix multiple editors.

But there is a problem here when using: `this.editor.config.get( 'blockToolbar' )` may not exist.
So we need to add a judgment here. If it doesn't exist, we won't deal with it.

### DIFF
--- a/src/toolbar/block/blocktoolbar.js
+++ b/src/toolbar/block/blocktoolbar.js
@@ -148,15 +148,19 @@ export default class BlockToolbar extends Plugin {
 	 * @inheritDoc
 	 */
 	afterInit() {
-		const factory = this.editor.ui.componentFactory;
 		const config = this.editor.config.get( 'blockToolbar' );
+		
+		if (config) {
+			const factory = this.editor.ui.componentFactory;
+			
+			this.toolbarView.fillFromConfig( config, factory );
 
-		this.toolbarView.fillFromConfig( config, factory );
-
-		// Hide panel before executing each button in the panel.
-		for ( const item of this.toolbarView.items ) {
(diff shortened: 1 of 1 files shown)


jodator:
Hi @Eusen, thanks for PR.

I see your problem and I'm not sure why we didn't define this config in the first place as in other plugins.

/cc @Reinmar - I think that we should use `config.define` always. AFAIR feature plugins do that so it might be better to add `editor.config.define( 'bockToolbar', [] );` because adding `if` to every place shouldn't be necessary.

Eusen:
I agree with you @jodator

Reinmar:
I guess we can define this prop to be empty. But then we need to make sure that the block toolbar button doesn't appear at all if this prop is left empty.

Please open a ticket for this so we can track the decision.

Eusen:
[#317](https://github.com/ckeditor/ckeditor5-utils/pull/317)

jodator:
Hi, @Eusen I'll close your PRs in favor of https://github.com/ckeditor/ckeditor5-ui/pull/534.

At the moment we will probably not change the API as proposed in https://github.com/ckeditor/ckeditor5-utils/pull/317. Also after checking how other features handles such cases I saw a different pattern:

```js
editor.config.get( 'foo' ) || [];
```
which should do the trick (not blow up the editor) and will be compatible with codebase.

Thanks for your input and sorry for longer response times.

PR #534, opened 2019-12-16
Make BlockToolbar work with empty configuration.

jodator:

Fix: Make `BlockToolbar` work with an empty configuration. Closes ckeditor/ckeditor5#5980.

---

### Additional information

* As discussed in https://github.com/ckeditor/ckeditor5-ui/pull/532 the block toolbar should work with empty toolbar configuration as other features use this pattern.

### DIFF
--- a/tests/toolbar/block/blocktoolbar.js
+++ b/tests/toolbar/block/blocktoolbar.js
@@ -55,6 +55,17 @@ describe( 'BlockToolbar', () => {
 		expect( BlockToolbar.pluginName ).to.equal( 'BlockToolbar' );
 	} );
 
+	it( 'should work with empty config', async () => {
+		// Remove default editor instance.
+		await editor.destroy();
+
+		expect( async () => {
+			editor = await ClassicTestEditor.create( element, {
+				plugins: [ BlockToolbar ]
+			} );
+		} ).to.not.throw();
+	} );
+
 	describe( 'child views', () => {
 		describe( 'panelView', () => {
 			it( 'should create a view instance', () => {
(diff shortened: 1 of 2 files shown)


tests/toolbar/block/blocktoolbar.js
```diff
@@ -55,6 +55,17 @@ describe( 'BlockToolbar', () => {
 		expect( BlockToolbar.pluginName ).to.equal( 'BlockToolbar' );
 	} );
 
+	it( 'should work with empty config', async () => {
```
Reinmar: I checked out `src/` to `master` and expected this test to fail, but it didn't. The other one did, but then why didn't this one? Maybe `expect()` doesn't handle async functions?

jodator: I'll check it then.

jodator: Oh... that was not wise for me :disappointed_relieved:.

With promises we have three choices:
1. Expecting to not fail not explicitly (anyway test must pass)

    ```js
    editor = await ClassicTestEditor.create( element, {
  	  plugins: [ BlockToolbar ]
    } );
    // Optional comment: The test must not throw or rename the test
    //  to "should not throw when empty config is provided"
    ```
2. Explicitly fail on error

    ```js
	try {
		editor = await ClassicTestEditor.create( element, {
			plugins: [ BlockToolbar ]
		} );
	} catch ( error ) {
		expect.fail( 'Error should not be thrown.' );
	}
    ```
3. Use chai-as-promised:

    ```js
	expect( async () => {
		editor = await ClassicTestEditor.create( element, {
			plugins: [ BlockToolbar ]
		} );
	} ).to.evantually.not.throw();
     ```

I'm for 1. But every other way works (3 needs update in the dev-tests to add chai-as-promised.

Reinmar: 1. is fine for me too.

PR #536, opened 2020-01-14
i/5597: Set balloon toolbar's max-width

panr:

Feature: The `BalloonToolbar` plugin should group items when its width is close to related editable's. Closes ckeditor/ckeditor5#5597. Closes ckeditor/ckeditor5#5501.

BREAKING CHANGE: The `BalloonToolbar` plugin groups overflowing items now by default. This can be disabled via the editor configuration by setting `config.balloonToolbar.shouldNotGroupWhenFull = true`.

[`BalloonPanelView.defaultPositions`](https://ckeditor.com/docs/ckeditor5/latest/api/module_ui_panel_balloon_balloonpanelview-BalloonPanelView.html#static-member-defaultPositions) has been extended with additional positions. Please refer to the documentation to learn more.

---

* [CI and constellation](https://github.com/ckeditor/ckeditor5/tree/i/5597)

Requires:
* [ ] https://github.com/ckeditor/ckeditor5-editor-inline/pull/65
* [ ] https://github.com/ckeditor/ckeditor5-theme-lark/pull/261
* [ ] https://github.com/ckeditor/ckeditor5-core/pull/211


tests/toolbar/toolbarview.js
```diff
@@ -70,6 +70,11 @@ describe( 'ToolbarView', () => {
 
 				toolbar.destroy();
 			} );
+
+			it( 'toolbar should group items by default', () => {
+				const toolbar = new ToolbarView( locale, { shouldGroupWhenFull: true } );
+				expect( toolbar.options.shouldGroupWhenFull ).to.be.true;
```
oleq: I'm not sure I get this test. You passed some options to the constructor and then you assert them, which has nothing to do with a "default" behavior as stated in the text description.

* If you wanted to add a test that checks if the options object is assigned as `ToolbarView#options`, then the previous text does exactly the same (so this one is obsolete).
* If you wanted to add a test to check what the default toolbar behavior of the toolbar is (i.e. no `shouldGroupWhenFull` in the constructor options means no grouping), then you should create a different test.

src/toolbar/balloon/balloontoolbar.js
```diff
@@ -44,6 +48,14 @@ export default class BalloonToolbar extends Plugin {
 	constructor( editor ) {
 		super( editor );
 
+		/**
+		 * A normalized `config.balloonToolbar` object.
+		 *
+		 * @type {Object}
+		 * @private
+		 */
+		this._balloonConfig = normalizeToolbarConfig( this.editor.config.get( 'balloonToolbar' ) );
```
oleq:
```suggestion
		this._balloonConfig = normalizeToolbarConfig( editor.config.get( 'balloonToolbar' ) );
```

src/toolbar/toolbarview.js
```diff
@@ -61,6 +61,14 @@ export default class ToolbarView extends View {
 		 */
 		this.set( 'ariaLabel', t( 'Editor toolbar' ) );
 
+		/**
+		 * Maximum width of the toolbar element.
+		 *
```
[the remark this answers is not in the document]
panr: 👍

tests/toolbar/balloon/balloontoolbar.js
```diff
@@ -344,6 +365,26 @@ describe( 'BalloonToolbar', () => {
 			balloonToolbar.show();
 			sinon.assert.calledOnce( balloonAddSpy );
 		} );
+
+		it( 'should set balloon toolbar max-width to half of the editable width, otherwise it can be wider then editor', done => {
+			const viewElement = editor.ui.view.editable.element;
+
+			setData( model, '<paragraph>b[ar]</paragraph>' );
+
+			expect( document.body.contains( viewElement ) ).to.be.true;
```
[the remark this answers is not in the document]
panr: Why should I use `utils/dom/global` only here? Any other test uses `/*global document... */`.
And I check the assertion here just because I like to know exactly where test breaks and why if so ¯\_(ツ)_/¯ 

If you really think that's redundant, I'll remove it ;-)

tests/toolbar/toolbarview.js
panr: You're right this test doesn't make sense 🙈

src/toolbar/balloon/balloontoolbar.js
```diff
@@ -125,6 +137,28 @@ export default class BalloonToolbar extends Plugin {
...
+					// We need to check if there's already the editable element in the DOM.
+					// Otherwise the `Rect` instance will complain that source (editableElement) is not available
+					// to obtain the element's geometry.
+					if ( !editableElement.ownerDocument.body.contains( editableElement ) ) {
+						return;
+					}
+
+					// In the balloon editor toolbar's max-width should be set to the 2/3 of the editable's width.
+					// It's a safe value, because at the moment we don't re-calculate it when position of the selection changes.
+					this.toolbarView.maxWidth = toPx( new Rect( editableElement ).width * 0.66 );
```
oleq: I checked this ratio and I came to the conclusion that it's unsafe for very small editables, e.g.:

image

Unfortunately, this case is totally valid, people may enable the editor on very narrow fields. 

So here's what I propose:
* instead 2/3, let's go with .9. This should still be safe and it will look better in cases like this where there's plenty of space left
image
* Let's have a sane min-width for the `BalloonToolbar` in CSS (so at least the heading dropdown is visible + maybe 2 or 3 buttons). As a custom CSS property so developers can override it if they want to. `min-width` should overrule `max-width` set by us in JS and this should work out great.

Reinmar: Can't we use the viewport width instead of editable width?

panr: I'm not sure about the viewport thing, because in some situations, someone can have the editor's width set to, for example: 60% of the viewport and then we want to make sure, that toolbar will not expand over the editor, covering some important stuff next to it.

oleq: I agree with @panr. The UI of the editor should be restricted to the editable whenever possible. This is "our territory" in the application and we should not expand beyond that.

panr: > instead 2/3, let's go with .9. This should still be safe and it will look better in cases like this where there's plenty of space left

I .9 is too much ATM. You can have that scenario: 

image

So if you want to set it to .9 we need to tweak anchoring behaviour as well.

oleq: We could extend default [`BalloonPanelView` positions](https://ckeditor.com/docs/ckeditor5/latest/api/module_ui_panel_balloon_balloonpanelview-BalloonPanelView.html#static-member-defaultPositions) and add intermediate ones and then [enable them in the `BalloonToolbar`](https://github.com/ckeditor/ckeditor5-ui/blob/master/src/toolbar/balloon/balloontoolbar.js#L285-L303).

[And we even have an issue for that :D](https://github.com/ckeditor/ckeditor5/issues/5501)

panr:
> Let's have a sane min-width for the BalloonToolbar in CSS (so at least the heading dropdown is visible + maybe 2 or 3 buttons). As a custom CSS property so developers can override it if they want to. min-width should overrule max-width set by us in JS and this should work out great.

Just added it. I set a default min-width of the balloon toolbar to `180px` to show only a heading dropdown and grouped items ([...]). Heading and 2-3 might be too much for super narrow editors (in sidebars or narrow panels).

image

src/toolbar/balloon/balloontoolbar.js
panr: Done, I added "middle-" states.

panr:
Sorry, I messed up with history merging fast forward `master` to this branch (stupid VSC) 😢

src/panel/balloon/balloonpanelview.js
```diff
@@ -450,14 +454,30 @@ BalloonPanelView._getOptimalPosition = getOptimalPosition;
  * 		+-----------------+
  * 		|     Balloon     |
  * 		+-----------------+
- * 		               V
- * 		          [ Target ]
+ * 		                V
+ * 		           [ Target ]
+ *
+ * * `northArrowSouthMiddleEast`
```
oleq: Let's sort all the positions within each group so they visually progress, for instance, from left to right.

Same in http://localhost:8125/ckeditor5-ui/tests/manual/panel/balloon/balloonpanelview.html.

src/panel/balloon/balloonpanelview.js
```diff
@@ -615,12 +716,24 @@ BalloonPanelView.defaultPositions = {
 		name: 'arrow_se'
 	} ),
 
+	northArrowSouthMiddleEast: ( targetRect, balloonRect ) => ( {
+		top: getNorthTop( targetRect, balloonRect ),
+		left: targetRect.left + targetRect.width / 2 - ( balloonRect.width * .75 ) + BalloonPanelView.arrowHorizontalOffset,
```
oleq: Probably unnecessary parentheses. The order of operators will take over.

src/toolbar/balloon/balloontoolbar.js
```diff
@@ -125,6 +148,26 @@ export default class BalloonToolbar extends Plugin {
 				this.show();
 			}
 		} );
+
+		if ( !this._balloonConfig.shouldNotGroupWhenFull ) {
+			this.listenTo( editor, 'ready', () => {
+				const editableElement = editor.ui.view.editable.element;
+
+				// Set toolbar's max-width on the initialization and update it on the editable resize.
+				this.widthObserver = new ResizeObserver( editableElement, () => {
```
oleq: This class has `this.resizeObserver` but no `#widthObserver`. 

BTW: if that passed in tests, they're probably not strong enough.

src/toolbar/balloon/balloontoolbar.js
```diff
@@ -60,6 +72,17 @@ export default class BalloonToolbar extends Plugin {
...
+		/**
+		 * An instance of the resize observer that helps dynamically determine the geometry of the toolbar
+		 * and manage items that do not fit into a single row.
+		 *
+		 * **Note:** Created in {@link #init}.
+		 *
+		 * @readonly
+		 * @member {module:utils/dom/getresizeobserver~ResizeObserver}
+		 */
+		this.resizeObserver = null;
```
oleq: Any reason to have it public? If not, then `@protected` or `@private` (if not used in tests), and preceded by an underscore.

src/toolbar/balloon/balloontoolbar.js
```diff
@@ -147,11 +189,27 @@ export default class BalloonToolbar extends Plugin {
...
+					'ck-toolbar_floating',
+					// When balloon toolbar contains `heading` plugin and the width of the editable is narrower than 200px,
+					// the balloon may fold too "agressively" and not show any of the buttons (only "three dots" button is visible
+					// to trigger dropdown menu with buttons that don't fit the toolbar width boundary).
+					//
+					// This is a workaround for the case. By default `ck-balloon-toolbar_min-width` class sets static `min-width` to 180px.
+					// The value matches a dynamic value (90% of the editable width) set by ResizeObserver's callback during initialization.
+					bind.if( 'maxWidth', 'ck-balloon-toolbar_min-width', value => {
+						if ( parseFloat( value ) <= 200 ) {
+							return balloonToolbarPlugins.find( plugin => plugin === 'heading' );
```
oleq: I think this heuristics is too náive. 

We don't know which plugins use dropdowns in the toolbar because not all of them come from CKSource. Developers can bring their custom plugins that register some dropdowns (named different than `'heading'`) and this optimization will fail.

Let's talk about this one-on-one because it might be tricky to figure out the right solution.

src/toolbar/toolbarview.js
```diff
@@ -61,6 +61,25 @@ export default class ToolbarView extends View {
 		 */
 		this.set( 'ariaLabel', t( 'Editor toolbar' ) );
 
+		/**
+		 * The maximum width of the toolbar element.
+		 *
+		 * Note:
+		 * You should not change the toolbar's max-width manually, because by default it's binded
+		 * with the editor's editable element's width and it will be applied automatically each time its value will change.
```
oleq: We can't write this. 

Toolbars are used in different places (for instance, highlight plugin drop-down or alignment drop-down) so `ToolbarView` is not limited to "editor toolbar" roles only. If anyone wants to use the toolbar anywhere else, that's fine and they can do with `#maxWidth` as they please.

It's a component. You should clearly state in docs what the property does and no more.

src/toolbar/toolbarview.js
```diff
@@ -61,6 +61,25 @@ export default class ToolbarView extends View {
...
+		 *
+		 * If you set `config.toolbar.shouldNotGroupWhenFull: true` and also set `ToolbarView#maxWidth`,
```
oleq: The `ToolbarView` class (as a UI component) knows nothing about editor config. It can be used in different ways in different contexts and should stay config-agnostic (also editor-agnostic). This is not the right place to write about `config.toolbar.shouldNotGroupWhenFull`. This probably should land as developer docs somewhere in editor (balloon, inline) implementations. 

You should **definitely** mention the relation with `module:ui/toolbar/toolbarview~ToolbarOptions#shouldGroupWhenFull` here, though.

src/toolbar/toolbarview.js
```diff
@@ -694,6 +717,21 @@ class DynamicGrouping {
 		this._updateGrouping();
 	}
 
+	/**
+	 * Enables the grouping functionality, just like {@link #_enableGroupingOnResize} but the difference is that
+	 * it listens to the {@link module:ui/toolbar/toolbarview~ToolbarView#maxWidth} changes.
+	 *
+	 * Note: Right now it works only with {@link module:editor-inline/inlineeditor~InlineEditor} and
```
oleq: Again, `ToolbarView` knows nothing about those packages and classes.

tests/toolbar/balloon/balloontoolbar.js
```diff
@@ -544,6 +615,36 @@ describe( 'BalloonToolbar', () => {
 		} );
 	} );
 
+	describe( 'has `Heading` plugin and editable width is less then 200px', () => {
```
oleq: As I mentioned in the class, this probably no the best solution to the problem.

src/panel/balloon/balloonpanelview.js
panr: Parentheses are just for readability purpose here ;-)

src/toolbar/balloon/balloontoolbar.js
panr: 😬

src/toolbar/balloon/balloontoolbar.js
panr: I made it `@protected` because I use it in `balloontoolbar` test.

src/toolbar/balloon/balloontoolbar.js
panr: Fair enough
ok-dale2

src/toolbar/toolbarview.js
panr: 👍

src/toolbar/toolbarview.js
panr: 👍

src/toolbar/toolbarview.js
panr: 👍

tests/toolbar/balloon/balloontoolbar.js
panr: Removed.

src/panel/balloon/balloonpanelview.js
panr: Done. I hope the changes are good enough.

oleq:
@ckeditor/qa-team Can you guys take a look at this PR constellation? Changes potentially affect all floating balloons in the editor.

Mgsy:
We've checked it and everything looks good 👍

PR #540, opened 2020-02-09
Added `tabindex="0"` to editable element to make it focusable in read-only mode

scofalik:

Other: Added `tabindex="0"` to editable element to make it focusable in read-only mode. Closes ckeditor/ckeditor5#6200.

---

### Additional information

I'd like this enhancement to be merged in iteration 29.

Three notes:

1. Tests. This will require thorough tests, some before merging and surely a lot during the testing phase. I didn't add too many tests in this PR, though. For unit tests, there's not a lot to test. When it comes to manual tests, we will have to keep this change in mind during the testing phase, so I am not sure if there's a need to make multiple/complex manual tests. I've added just one manual test that allows checking how the focus on <kbd>Tab</kbd> changes and how the editor behaves in read-only mode.

2. There's one important note about the editor's behavior in the read-only mode after this change. Balloon contextual toolbars are now visible. For example, when you click an image in the read-only mode, the toolbar for the image will appear. It will be all-disabled, but it will appear. I guess this is something that we might want to discuss.

3. I've added `tabindex="0"` not only for the read-only mode but also for the normal mode. In theory, this should not have an impact on the editor's behavior in the normal mode. But if you are afraid, I can change that and bind `tabindex="0"` to the read-only mode state.

If you want, I can try to come up with other things in the manual test (or expand the description) but again, since this is a change can affect multiple features in the read-only mode, I think that we should simply keep this change in mind during the testing phase.


tests/manual/tickets/6200/1.js
```diff
@@ -0,0 +1,74 @@
...
+BalloonEditor
+	.create( document.querySelector( '#editor-balloon' ), EDITOR_CONFIG )
+	.then( editor => {
+		window.balloonEditor = editor;
+	} )
+	.catch( err => {
+		console.error( err.stack );
+	} );
+
+document.querySelector( '#btn-readonly' ).addEventListener( 'click', () => {
```
Reinmar: Looks like a problem with indentation here?

oleq: Why?

PR #542, opened 2020-02-17
In ContextualBalloon#remove cache stack id before currently shown view is changed

scofalik:

Internal: In `ContextualBalloon#remove` cache stack id before currently shown view is changed.

---

### Additional information

This change prevents an editor crash when `ContextualBalloon#remove` fires itself with the same view (as an effect of a chain of events).

See linked CF issue.

Following interaction happens in the comments feature:

1. Selection is in two comments markers, inside the inner comment, the balloon has two views.
2. Selection is moved to the outer marker, `remove()` is fired.
3. Inside `remove()` a couple of things happen that lead to firing `remove()` again with the same view.
4. The view is removed in the "second" `remove()`.
5. But then it crashes when the execution flow goes back to the original `remove()`.

This fix prevents crashing in the original `remove()` by saving `stackId`.

### DIFF
--- a/src/panel/balloon/contextualballoon.js
+++ b/src/panel/balloon/contextualballoon.js
@@ -258,6 +258,7 @@ export default class ContextualBalloon extends Plugin {
 		}
 
 		const stack = this._viewToStack.get( view );
+		const stackId = this._getStackId( stack );
 
 		if ( this._singleViewMode && this.visibleView === view ) {
 			this._singleViewMode = false;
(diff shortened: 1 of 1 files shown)


Reinmar:
cc @ckeditor/qa-team This isn't the easiest things to review at this stage and it's hard to tell how it can affect the editor balloon system, so a confirmation from you that it doesn't break anything would be great.

scofalik:
We decided to implement the fix for this issue on the CF side.

PR #545, opened 2020-02-27
Added support for initializing ViewCollection and BodyCollection items using a constructor

mlewand:

Feature: Added support for initializing `ViewCollection` and `BodyCollection` items using a constructor. See ckeditor/ckeditor5#6319.

The `View.createCollection()` method also accepts an iterator of views.

MAJOR BREAKING CHANGE: `ViewCollection` no longer has the `locale` property.

MAJOR BREAKING CHANGE: `ViewCollection`'s constructor no longer takes the `locale` as a parameter.

---

### Additional information

This PR was mostly based on #524 with some notable changes:

*   Back then we could have removed locale from all collections. Today locale is required by one `Collection` subclass - `BodyCollection` (it uses it to determine text direction). So it is only BodyCollection that contains this property.
*   `ViewCollection` / `BodyCollection` accepts any iterable of views, not just an array.

This PR requires a PR from utils to be working: ckeditor/ckeditor5-utils#326

After it's done, the `master` branch should be merged into #524 PR so that the diff there is much smaller and easier to review/finish it.


mlewand:
Not sure what's up with the code coverage as I'm having 100% CC both, when running UI module alone and all the tests (here technically it's 99.9% due to #6326):

UI module:

```plaintext
Chrome 80.0.3987 (Windows 10.0.0): Executed 868 of 868 SUCCESS (10.824 secs / 1.831 secs)

=============================== Coverage summary ===============================
Statements   : 100% ( 1575/1575 )
Branches     : 100% ( 604/604 )
Functions    : 100% ( 450/450 )
Lines        : 100% ( 1549/1549 )
================================================================================
```

all:

```plaintext
Chrome 80.0.3987 (Windows 10.0.0): Executed 13317 of 13339 (skipped 22) SUCCESS (6 mins 4.766 secs / 2 mins 5.142 secs)

=============================== Coverage summary ===============================
Statements   : 100% ( 18726/18726 )
Branches     : 99.99% ( 9621/9622 )
Functions    : 100% ( 4190/4190 )
Lines        : 100% ( 18447/18447 )
================================================================================
```

src/view.js
```diff
@@ -276,10 +276,11 @@ export default class View {
 	 *		// <p><child#element></p>
 	 *		view.items.add( child );
 	 *
+	 * @param {Iterable.<module:ui/view~View>} [views] Initial views of the collection.
 	 * @returns {module:ui/viewcollection~ViewCollection} A new collection of view instances.
 	 */
-	createCollection() {
-		const collection = new ViewCollection();
+	createCollection( views ) {
```
oleq: Missing docs for the usage with `views`.

src/viewcollection.js
```diff
@@ -78,14 +72,6 @@ export default class ViewCollection extends Collection {
 			}
 		} );
 
-		/**
-		 * The {@link module:core/editor/editor~Editor#locale editor's locale} instance.
-		 * See the view {@link module:ui/view~View#locale locale} property.
-		 *
-		 * @member {module:utils/locale~Locale}
-		 */
-		this.locale = locale;
```
oleq: Was it really obsolete? Because it does not feel like the scope of the PR. OTOH it makes the arguments discovery harder in the `constructor()`, that's why I'm asking.

src/viewcollection.js
```diff
@@ -194,6 +185,30 @@ export default class ViewCollection extends Collection {
...
+	 * @param {module:ui/view~View} view A new view added to the collection.
+	 * @param {Number} [index] An index the view holds in the collection. When not specified,
+	 * the view is added at the end.
+	 */
+	_renderViewIntoCollectionParent( view, index ) {
+		if ( !view.isRendered ) {
+			view.render();
+		}
+
+		if ( view.element && this._parentElement ) {
```
oleq: How can a `View` have no `#element` if it's been just rendered in :204?

src/editorui/bodycollection.js
```diff
@@ -33,6 +33,24 @@ import createElement from '@ckeditor/ckeditor5-utils/src/dom/createelement';
...
  */
 export default class BodyCollection extends ViewCollection {
+	/**
+	 * Creates a new instance of the {@link module:ui/editorui/bodycollection~BodyCollection}.
+	 *
+	 * @param {module:utils/locale~Locale} locale The {@link module:core/editor/editor~Editor editor's locale} instance.
+	 * @param {Iterable.<module:ui/view~View>} [initialItems] The initial items of the collection.
+	 */
+	constructor( locale, initialItems = [] ) {
+		super( initialItems );
```
oleq: I see no test that verifies that `initialItems` is passed to the `ViewCollection#constructor`.

oleq:
Also, we should review all places that do `createCollection` or `new ViewCollection` and use the new API if possible. This may require some understanding of what's going on there, though.

src/view.js
mlewand: Addressed in e470ac51.

src/viewcollection.js
mlewand: That's [unmodified original code from the `master` branch](https://github.com/ckeditor/ckeditor5-ui/blob/5d86638d48d62680188f863a3a052a7d53c8e30d/src/viewcollection.js#L65-L71), it has just been extracted to a named method - and it wasn't really my intention to tackle issues not related with the issue.

Giving it a quick look the element could be unavailable if [the view has no template](https://github.com/ckeditor/ckeditor5-ui/blob/5d86638d48d62680188f863a3a052a7d53c8e30d/src/viewcollection.js#L65-L71) sounds like an edge case, bot something that might happen? Eventually this refactoring might be extracted to a follow up in order not to block this PR.

src/viewcollection.js
mlewand: As you pointed out back in #524 - locale in `ViewCollection` has no usage in our codebase. However sine then `BodyColleciton` subclass has been introduced which does use it.

More on that in the PR main description:

> Back then we could have removed locale from all collections. Today locale is required by one `Collection` subclass - `BodyCollection` (it uses it to determine text direction). So it is only BodyCollection that contains this property. 

> Because it does not feel like the scope of the PR

I do agree on the fact that it goes outside of the scope, I'm fine with extracting this - let me know if clarification are good enough or you'd prefer to extract it to a separate issue.

src/editorui/bodycollection.js
mlewand: Addressed in c71c323

mlewand:
> Also, we should review all places that do `createCollection` or `new ViewCollection` and use the new API if possible. This may require some understanding of what's going on there, though.

There will be quite a few of these and indeed it will involve checking surrounding code, so I'd like to do that once there are no other remaining points for this PR.

PR #546, opened 2020-03-17
I/6109: Add FormHeaderView component

panr:

Feature: Add FormHeaderView UI component. Closes ckeditor/ckeditor5#6109.


src/formheader/formheaderview.js
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * @module ui/formheader/formheaderview
+ */
+
+import View from '../view';
+
+import '../../theme/components/formheader/formheader.css';
+
+/**
+ * The class representing a form header view, used i.a.
```
jodator: Could you add some more meaningful description? Ie why is this class in the first place what's its purpose and how/when it can be used (probably one sentence would be enough). Where it is used is rather not helpful.

src/formheader/formheaderview.js
```diff
@@ -12,10 +12,18 @@ import View from '../view';
...
 
 /**
- * The class representing a form header view, used i.a.
+ * The class component representing a form header view. It should be used in more advanced forms to
+ * describe the main purpose of the form.
+ *
+ * By default the component contains bolded label view which have to be set. The label is usually a short (at most 3-word) string.
+ * The component can also be extended by any other elements, like: icons, dropdowns etc.
+ *
+ * It is used i.a.
```
jodator:
```suggestion
 * The class component representing a form header view. It should be used in more advanced forms to
 * describe the main purpose of the form.
 *
 * By default the component contains a bolded label view that has to be set. The label is usually a short (at most 3-word) string.
 * The component can also be extended by any other elements, like: icons, dropdowns, etc.
 *
 * It is used i.a.
```

panr: Wow, `_have_` — what a "siara" 😬Glad all our code is private 😅

PR #547, opened 2020-03-20
I/6110: Change name LabeledView -> LabeledFieldView

panr:

Other: Changed the name of the `LabeledView` component to the `LabeledFieldView`. See ckeditor/ckeditor5#6110.

Related PR: https://github.com/ckeditor/ckeditor5-theme-lark/pull/274

--
### BREAKING CHANGES:

- From now the instance of a labeled component's view is available through `LabeledFieldView#field`. It replaced `LabeledView#view`.


panr:
FYI: CI breaks due to the fact that I use `console.warn`.

panr:
I'm struggling with CI. Tried almost everything... What I'm missing? 🤔

src/labeledfieldview/labeledfieldview.js
```diff
@@ -175,7 +175,7 @@ export default class LabeledView extends View {
 			},
 			children: [
 				this.labelView,
-				this.view,
+				this.field,
```
jodator: Should be `this.fieldView` as it hurts when you read those 3 in a row ;) We should keep property naming consinstent.

panr: Hmm... I had a talk with @Reinmar and @oleq about the naming and the suggestion was to use `this.field`, but right now I'm not sure... `this.fieldView` looks/feels better IMHO. (sooooo many changes all over again 😵)

jodator: OH I didn't know about that :( But looking at other props `fieldView` looks better :/). You can always rename others to match  :P

Reinmar: fieldView :+1:

panr: OK, I'm on it.

panr:
@jodator all done ✅🤞

PR #550, opened 2020-03-30
I/6507: Clean up the form header component

panr:

Other: Cleaned up the component docs and code. Closes ckeditor/ckeditor5#6507.


panr:
Fair enough, I’ll fix this 👍🏻

panr:
@oleq all right, fixed it as it should be from the beginning ;-)

PR #552, opened 2020-04-03
i/6087: Created the "Deep dive into focus tracking" guide.

oleq:

Docs: Created the "Deep dive into focus tracking" guide. Closes ckeditor/ckeditor5#6087.

---

### Additional information

**Merge together with**: 
* https://github.com/ckeditor/ckeditor5/pull/6551
* https://github.com/ckeditor/ckeditor5-utils/pull/330

Follow-ups:
* https://github.com/ckeditor/ckeditor5/issues/6548


docs/framework/guides/deep-dive/focus-tracking.md
```diff
@@ -0,0 +1,731 @@
...
+
+## Focus in the editor engine
+
+<info-box>
+	Keep in mind that information in this chapter concern the editor **engine layer only**. To learn about focus in the user interface, check out the [next chapter](#focus-in-the-editor-ui) of this guide.
+</info-box>
+
+The main editable area of CKEditor can be focused thanks to the [`contenteditable`](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Editable_content) DOM attribute. This attribute tells the web browser that a web page element can be edited like any other text field, which also means it must be able to receive focus.
+
+Each root of the editing view has the `contenteditable` attribute. The editing view uses the {@link module:utils/focustracker~FocusTracker `FocusTracker`} (learn more about {@link framework/guides/architecture/editing-engine#observers view observers}) to track focus in editables by listening to native DOM `focus` and `blur` events coming from them.
```
Reinmar, line 43: Wasn't this supposed to be about `FocusObserver`?

docs/framework/guides/deep-dive/focus-tracking.md
```diff
@@ -0,0 +1,731 @@
...
+
+<info-box>
+	Already confused? Take a look at our {@link framework/guides/architecture/editing-engine#observers "Editing engine"} guide that will explain what the editing view, editables and other building blocks of CKEditor are.
+</info-box>
+
+### Checking if the view document is focused
+
+You can check if the view document is focused using its {@link framework/guides/deep-dive/observables observable} `isFocused` property. Create an editor instance and execute the following code:
```
Reinmar, line 51: Missing link to `isFocused`.

docs/framework/guides/deep-dive/focus-tracking.md
```diff
@@ -0,0 +1,731 @@
...
+Say, for instance, your application wants to focus an editable area of CKEditor when a certain action is executed (e.g. a button is clicked). To do that, use the {@link module:engine/view/view~View#focus `focus()`} method of the editing view:
+
+```js
+editor.editing.view.focus();
+```
+
+This snippet focuses an editable that has the selection. If the editor has not been focused yet, this will focus the very first editable. If an editor has multiple editing roots and the user was editing content, focus will be brought back where the user left off.
+
+<info-box>
+	Focusing an editor does not change its selection. If you want to focus an editor and move the caret to a specific position, you should call `editor.editing.view.focus()` first and then use the {@link framework/guides/architecture/editing-engine#model model writer} to change the selection.
```
Reinmar, line 82: I'd link to setSelection directly.

docs/framework/guides/deep-dive/focus-tracking.md
```diff
@@ -0,0 +1,731 @@
...
+</info-box>
+
+If you want to focus a particular editable of your editor (not just the first one), you should take a closer look at the {@link module:engine/view/domconverter~DomConverter#focus `DomConverter#focus()`} method that will do that for you.
```
Reinmar, line 85: I don't think this is necessary for anyone to do that. Once you set the model selection and focus the editing view, it should take care of focusing the correct editable element.

docs/framework/guides/deep-dive/focus-tracking.md
```diff
@@ -0,0 +1,731 @@
...
+
+## Focus in the editor UI
+
+If you read the [previous chapter](#focus-in-the-editor-engine) of this guide you should know that there is already a layer responsible for tracking focus working at the engine–level. But while that layer is only concerned by the focus in editor editables, due to the {@link framework/guides/architecture/intro modular nature of the editor framework}, it is oblivious of the user interface. This makes sense, for instance, because it is possible to create a fully–functioning editor without the UI and there should be a minimal API provided by the editor engine that allows that.
```
Reinmar, line 89: "engine level"?

Reinmar: "editor editables" is a wording we used in CKE4 and I rather prefer less hermetic names like "editable roots of the editor"

Reinmar: "editor editables" is a wording we used in CKE4 and I rather prefer less hermetic names like "editable roots of the editor"

Reinmar: could you replace "oblivious" with something I can understand? :D

Reinmar: As for the last sentence – perhaps it could be simplify by starting it with "Thanks to that it is possible..."?

docs/framework/guides/deep-dive/focus-tracking.md
```diff
@@ -0,0 +1,731 @@
...
+
+As for the user interface of CKEditor, it is a composition of multiple components {@link framework/guides/architecture/ui-library#view-collections-and-the-ui-tree organized as a tree}. This tree determines not only the logical structure of the UI (a toolbar has a dropdown, a dropdown has a button, a button has an icon, etc.) but also its behavior, and that includes tracking and maintaining focus as users navigate and interact with various pieces of the interface.
```
Reinmar, line 91: Singular user reads more natural to me:

```suggestion
As for the user interface of CKEditor, it is a composition of multiple components {@link framework/guides/architecture/ui-library#view-collections-and-the-ui-tree organized as a tree}. This tree determines not only the logical structure of the UI (a toolbar has a dropdown, a dropdown has a button, a button has an icon, etc.) but also its behavior, and that includes tracking and maintaining focus as the user navigates and interacts with various pieces of the interface.
```

docs/framework/guides/deep-dive/focus-tracking.md
```diff
@@ -0,0 +1,731 @@
...
+Most of components have [focus trackers](#using-the-focustracker-class) to keep up with the focus inside of them. Some components that host more children also use [focus cyclers](#using-the-focuscycler-class) and [keystroke handlers](#using-the-keystrokehandler-class) to allow users to navigate across them. You can learn how to use them in the later chapters of this guide:
+
+- ["Implementing focusable UI components"](#implementing-focusable-ui-components),
+- ["Using the `FocusTracker` class"](#using-the-focustracker-class),
+- ["Using the `KeystrokeHandler` class"](#using-the-keystrokehandler-class),
+- ["Using the `FocusCycler` class"](#using-the-focuscycler-class).
+
+### Implementing focusable UI components
+
+Any UI {@link framework/guides/architecture/ui-library#views view} can be focusable. To become one, a view must implement the `focus()` method that focuses the DOM {@link module:ui/view~View#element  element} and the `tabindex="-1"` attribute set on the element that prevents the native navigation using the keyboard (which should be handled by the [focus cycler](#using-the-focuscycler-class) on the parent–level):
```
Reinmar, line 185: I'd add an infobox here explaining that tabindex=-1 is a native concept and what it does because it may seem like our thing.

docs/framework/guides/deep-dive/focus-tracking.md
```diff
@@ -0,0 +1,731 @@
...
+	}
+
+	render() {
+		super.render();
+
+		// ...
+
+		// Start listening for the keystrokes coming from #element, which will allow
+		// the focus cycler to handle the keyboard navigation.
+		this.keystrokes.listenTo( this.element );
```
Reinmar, line 350: I wonder if we don't need a destroy() here. Theoretically there's no need to unbind the keystroke handler from this element as both get garbage collected. But that happens only if the element is removed from the DOM tree. Do we assume here that this is always true?

Reinmar:
The guide is fantastic. I think it's by far the best one we have :)

I think there's only one thing that I miss – a clear explanation that there's this main editor focus tracker and that all focusable roots are registered here and a list of these (toolbar, editables, etc).

And then a word that we consider the editor having focus as long as any of these items have focus.

Theoretically most of these is explained in `focus-tracking.html#focus-in-the-editor-ui` but after reading this guide I can't recall this being mentioned, so perhaps some sort of summary is needed. Perhaps this would be a good place:

Reinmar:
Haha :D I've had one last "Conclusions" part to read before I wrote the above comment and I can see that it seems to clarify some of those things. Still, I'd have it somewhere earlier too.

docs/framework/guides/deep-dive/focus-tracking.md
oleq: A nice catch. Here's the follow-up https://github.com/ckeditor/ckeditor5/issues/6561.

docs/framework/guides/deep-dive/focus-tracking.md
oleq: No, why do you think it should?

oleq:
> Haha :D I've had one last "Conclusions" part to read before I wrote the above comment and I can see that it seems to clarify some of those things. Still, I'd have it somewhere earlier too.

TBH, I couldn't find any good place to explain it earlier. Do you have any idea where could it be?

Reinmar:
> TBH, I couldn't find any good place to explain it earlier. Do you have any idea where could it be?

I think you could just add this additional info as another points in the list I mentioned in https://github.com/ckeditor/ckeditor5-ui/pull/552#issuecomment-608721777.

oleq:
> I think you could just add this additional info as another points in the list I mentioned in [#552 (comment)](https://github.com/ckeditor/ckeditor5-ui/pull/552#issuecomment-608721777).

Done. They only unresolved discussion is https://github.com/ckeditor/ckeditor5-ui/pull/552#discussion_r403106033.

docs/framework/guides/deep-dive/focus-tracking.md
Reinmar: Because you link to view observers, so the thing you're talking about is FocusObserver, not FocusTracker. The engine view definitely uses focus observer.

oleq: You're right, my bad.

PR #554, opened 2020-04-08
i/6449: Make block toolbar groupable

### LINKED ISSUE #6449: Block toolbar in Balloon Block Editor build doesn't group items
## Provide a description of the task

While fixing the issue related to toolbar grouping in Balloon and Inline Editor I forgot to add support for this to `balloon block toolbar`, so in the Balloon Block Editor build (https://ckeditor.com/docs/ckeditor5/latest/examples/builds/balloon-block-editor.html) we have something like this in the smaller viewports.

image

In this case, the best solution (probably) is to bind the `maxWidth` value to the `window` width, not the `editable` as we did in balloon and inline toolbar.

## 📃 Other details

Issue is related to https://github.com/ckeditor/ckeditor5/issues/5597

### LINKED ISSUE #6570: BlockToolbar looses its position and width
## 📝 Provide detailed reproduction steps (if any)

💡This issue has been created before the code got merged and it exists only locally in situation where `BlockToolbar` has implemented `shouldGroupWhenFull`. 

The issue is needed only as a reference to possibly existing bug and it is related to https://github.com/ckeditor/ckeditor5/issues/6449.

1. Allow `BlockToolbar` to group items.
2. Show toolbar panel.
3. Hide it.
4. Show it again.

### ✔️ Expected result

Toolbar panel should be anchored correctly to the block toolbar button each time.

### ❌ Actual result

Toolbar panel is anchored correctly only the first time you show it. It resets the position and width only when view has been resized.

issue

### LINKED ISSUE #6575: ToolbarView should handle items grouping correctly when invisible
## 📝 Provide detailed reproduction steps (if any)

1. The grouping should never happen if the toolbar is invisible. Because it always leads to all items groped and weird things when other features (like positioning algorithm) when they see the toolbar in such state

image

2. The grouping that was triggered by changing `ToolbarView#maxWidth` when toolbar is invisible should be queued so the next time the toolbar shows up (`ResizeObserver` kicks in automatically), the items are re-grouped. This is a problem when `maxWidth` increased because there's no pressure to ungroup the items when the toolbar showed up.

Related to https://github.com/ckeditor/ckeditor5/issues/6449.

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

panr:

Feature: The `BlockToolbar` should group items when there is no place to show them all. Closes ckeditor/ckeditor5#6449. Closes ckeditor/ckeditor5#6575. Closes  ckeditor/ckeditor5#6570.

Bulletproofed the `ToolbarView#maxWidth` and items grouping when the toolbar is invisible.

---

### Additional information

This PR closes ckeditor/ckeditor5#6449 but it is related also to the issue ckeditor/ckeditor5/issues/6570.


src/toolbar/block/blocktoolbar.js
```diff
@@ -388,6 +456,23 @@ export default class BlockToolbar extends Plugin {
 		this.buttonView.top = position.top;
 		this.buttonView.left = position.left;
 	}
+
+	/**
+	 * Sets block {@link #toolbarView} max-width, based on
```
oleq, line 461:
```suggestion
	 * Gets the {@link #toolbarView} max-width, based on
```

src/toolbar/block/blocktoolbar.js
```diff
@@ -388,6 +456,23 @@ export default class BlockToolbar extends Plugin {
...
+	 * editable width plus distance between furthest edge of the {@link #buttonView} and the editable.
```
oleq, line 462:
```suggestion
	 * editable width plus distance between farthest edge of the {@link #buttonView} and the editable.
```

src/toolbar/block/blocktoolbar.js
```diff
@@ -388,6 +456,23 @@ export default class BlockToolbar extends Plugin {
...
+	 *
+	 * @private
+	 * @returns {String} maxWidth A maximum width that toolbar can have, sets in pixels.
```
oleq, line 465:
```suggestion
	 * @returns {String} maxWidth A maximum width that toolbar can have, in pixels.
```

oleq (changes requested):
We're only missing docs about enabling/disabling grouping
* here https://ckeditor.com/docs/ckeditor5/latest/features/blocktoolbar.html#configuration
* and here https://ckeditor.com/docs/ckeditor5/latest/api/module_core_editor_editorconfig-EditorConfig.html#member-blockToolbar

src/toolbar/block/blocktoolbar.js
panr: That's interesting... 🤔

